PR #4, opened 2018-05-26
Add requirements.txt

silasary:


ZeldaZach:
Is this how a python package lists dependencies outside of normal python scope?

AridTag:
[seems that way](https://pip.pypa.io/en/latest/user_guide/#requirements-files)

silasary:
Yup. It's the way you share pypy dependencies with other developers/users.

>

ZeldaZach:
Sounds good to me. Are we sure these are the only dependencies (so far)?

silasary:
So far, yes.
To make it, I created an empty virtual environment, then added packages until it started working.

PR #16, opened 2018-05-29
Fix almost all mypy errors

Daenyth:
There's a few left; run `tox -e mypy`

### DIFF
--- a/mtgjson4/mtg_builder.py
+++ b/mtgjson4/mtg_builder.py
@@ -7,7 +7,9 @@ import json
 from mtgjson4 import mtg_global, mtg_http, mtg_parse, mtg_storage, corrections
 import os
 import pathlib
-from typing import Any, Dict, List, Optional, Union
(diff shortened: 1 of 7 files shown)


mtgjson4/mtg_builder.py
```diff
@@ -86,8 +88,8 @@ def determine_layout_and_div_name(
         card_info['cmc'] = mtg_parse.parse_card_cmc(soup_oracle, div_name)
 
         # Get other side's name for the user
-        has_other, card_other_name = mtg_parse.parse_card_other_name(soup_oracle, div_name, card_layout)
-        if has_other:
+        card_other_name = mtg_parse.parse_card_other_name(soup_oracle, div_name, card_layout)
+        if card_other_name is not None:
```
Daenyth, line 92: Note this pattern is much safer; mypy is aware of `is not None` checks and will promote `Optional[A]` to `A` inside the `if` branch

PR #24, opened 2018-06-24
Attempt to ensure params are const and not modified

ZeldaZach:
WIP PR

### DIFF
--- a/mtgjson4/mtg_builder.py
+++ b/mtgjson4/mtg_builder.py
@@ -142,7 +142,7 @@ class MTGJSON:
 
     async def build_main_part(self,
                               set_name: List[str],
-                              card_info: mtg_global.CardDescription,
+                              card_mid: int,
                               other_cards_holder: Optional[List[object]],
                               second_card: bool = False) -> None:
         """
@@ -152,9 +152,11 @@ class MTGJSON:
         This builder only builds the main page of Gatherer.
         """
 
+        # This will be the holder that is returned
+        card_info = dict()
+
         # Parse web page so we can gather all data from it
-        card_mid = card_info['multiverseid']
-        soup_oracle = await self.get_card_html(int(str(card_mid)))
+        soup_oracle = await self.get_card_html(card_mid)
 
         card_layout, div_name, alt_div_name, add_other_card = self.determine_layout_and_div_name(
             soup_oracle, second_card)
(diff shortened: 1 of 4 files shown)


mtgjson4/mtg_builder.py
```diff
@@ -263,7 +265,9 @@ def determine_layout_and_div_name(soup: bs4.BeautifulSoup,
         if c_variations:
             card_info['variations'] = c_variations
 
-    async def build_legalities_part(self, card_mid: int, card_info: mtg_global.CardDescription) -> None:
+        return card_info
+
+    async def build_legalities_part(self, card_mid: int) -> None:
```
Daenyth: Need to adjust the return type here.

mtgjson4/mtg_builder.py
```diff
@@ -333,6 +348,14 @@ def determine_layout_and_div_name(soup: bs4.BeautifulSoup,
         if c_original_text:
             card_info['originalText'] = c_original_text
 
+        return card_info
+
+    @staticmethod
+    def combine_all_parts_into_a_card(main_parts, original_parts, legal_parts, foreign_parts) -> Dict[str, Any]:
+        dict_builder1 = {**main_parts, **original_parts}
+        dict_builder2 = {**legal_parts, **foreign_parts}
+        return {**dict_builder1, **dict_builder2}
```
Daenyth: why not `return {**main_parts, **original_parts, **legal_parts, **foreign_parts}`?

You should also `-> CardDescription` on this method so you can push the return type up the stack from there

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = yapf-check,mypy,lint,isort-check,unit
+envlist = yapf-inplace,mypy,lint
```
Daenyth, line 2: no `unit` when you run `tox`?

Daenyth (approved):
Looks like a step in the right direction

mtgjson4/mtg_builder.py
ZeldaZach: Was based on SO and didn't know if I could do that, thanks for the tip

tox.ini
ZeldaZach: Just for now, didn't see unit doing anything

PR #37, opened 2018-08-03
Pathlib + Py3.7 + NoEval

### LINKED ISSUE #34: Consistent File Access (using pathlib.Path instead of os.path)
We use both `os` and `pathlib` libraries for reading/writing files. I'd like to fix up the code to use either or. We can have a discussion on which is better here, but it seems a bit weird having two libs in that do the same thing

ZeldaZach:
This PR moves us away from os.path and to using pathlib.Path. In addition, we no longer use EVAL for our cached files, update us to python 3.7 over 3.6.5, and fixes a few minor spelling errors.

Fix #34


silasary:
You also need to update `.travis.yml` to 3.7

tooomm:
Same for the readme file. It also needs an update to **3.7** ;)

ZeldaZach:
Addressed @silasary @tooomm

silasary (approved):
Looks good to me.

Issue #35, opened 2018-07-09
Set Config Standardization


ZeldaZach:
The formatting of the set_configs isn't consistent, and I'd like to change that. We should specify what field(s) are required/optional in a schema/documentation.

My initial thought is
- set name
- set code
- release date
- border
- type
- booster pack contents (TODO: Needs to be reworked)
- languages available in 
- scryfall key (if necessary)

but I'm open to suggestions

I know @ancestral started on this a while ago, but I never heard back

tooomm:
For later, maybe we can have travis check that all fields we want/need are provided for new entries.

Like with the changelog check introduced in https://github.com/mtgjson/mtgjson/pull/381 or the json check+lint from https://github.com/mtgjson/mtgjson/pull/431.

ZeldaZach:
Moved away from this, using SF's version instead.

Missing booster fields, but we'll consider alternatives in the future.

PR #38, opened 2018-08-03
C18 metadata

taw:
[Commander 2018](https://magic.wizards.com/en/products/Commander-2018) is coming out.

It seems that some but not all sets have extra fields like translations etc., just added same fields as C17 had.


ZeldaZach:
Relates to #35

PR #89, opened 2018-10-17
Create version files

ZeldaZach:
I'm not sure if I'm happy with my implementation or idea of how to generate these files. Feedback highly welcomed


ZeldaZach:
Yeah I'm not happy with this.. Will reconsider what I want to do in time for 4.1

PR #99, opened 2018-10-23
Configure travis and pre-commit to test, lint, and check.

gwax:
Add pre-commit hooks

Add mypy checking

Add a donothing test so that pytest does not fail for lack of tests

Tweaks to try to get travis working.

Add eclint checking

Update pylintrc

Fix isort config

Pre-commit fix all the things

travis Python -> 3.7.1

Use xenial to build against 3.7

Seems to work without npm install

black some more

isort fix

isort config tweak

eclint config fixes

eclint fixes

Fix editorconfig


silasary (approved):

silasary:
Looks good to me 👍

silasary:
And yeah, we can remove Codacy if this gets merged in.

.editorconfig
```diff
@@ -1,8 +1,15 @@
+root = true
+
 [*]
-trim_trailing_whitespace = true
+end_of_line = lf
```
silasary: I'd rather this line not exist, as it's something that should be handled by your git config, and causes issues for windows users (like myself) who use git with `core.autocrlf = true` (this is enabled by default)

gwax: With `autocrlf` in place, where does this impact you?

If my understanding is correct, git will account for this before pre-commit checks things and it will be stored in the repo as `lf` so it will pass CI.

Is the problem that git with change `lf` to `crlf` for you but your editor will interpret EditorConfig to convert it back?

I'm not opposed to removing this but I would prefer if this check existed for the CI-side `eclint` check.

silasary: So with the editorconfig like this, you have the following workflow:

1. Check out a commit.  All files are CRLF.
2. do some work.  Every time your IDE saves a file, it changes the lines in that file to LF.
3. go to commit. Git complains at you that your line endings are inconsistent.  
```
warning: LF will be replaced by CRLF in __init__.py.
The file will have its original line endings in your working directory.
```
Depending on settings and whether or not you're using a graphical git client, this may be presented as an error.

It'll also cause merge conflicts if you then try to pull.  Easy conflicts that don't require hand-merging, but still merge conflicts.

The editorconfig [documentation](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#end_of_line) recommends leaving this unset.

gwax: Done. We can always revisit later.

PR #100, opened 2018-10-27
Rewrite Gatherer and better separate some interfaces

gwax:
This PR rewrites the gatherer parsing to dig into individual card entries ("rightCol") and then disassemble their `Label`/`Value` lists. In the process of doing so, I have fixed `originalText` and `originalType` for double faced cards (diff ISD.json against master build for comparison).

Tests have been added for the Gatherer parsing.

There is further separation of the data providers and card building.


ZeldaZach:
Amazing job, just needed the requirement to function :)

gwax:
@ZeldaZach Thanks for the merge but I'm not sure why you added `contextvars` to the requirements, it's in the standard library: https://docs.python.org/3/library/contextvars.html

ZeldaZach:
PyCharm told me the dependency was missing and to add it to the requirements.txt document... Weird how it didn't pick up on that then

PR #104, opened 2018-10-30
Address split/meld card issues

### LINKED ISSUE #90: Names field on meld cards is out of order
The "names" array on all multi-part cards is the only way to tell which part is which, and as such its order is important. For meld cards, the order should be: top half, bottom half, melded card. This is the intuitive order, and is the way it was done in v3. Currently however the names arrays on the meld cards are in a seemingly random order, not even consistent among themselves.

### LINKED ISSUE #91: The names field is severely messed up on ~70% of multi-part cards.
Looks like the recent change to the names array has caused some... problems. Budoka Gardener has the array of ["Budoka Gardener // Dokai, Weaver of Life"], Start // Finish has ["Start // Finish", "Warrior"], and Arlinn Kord has ["Arlinn Kord // Arlinn, Embraced by the Moon", "Wolf", "Arlinn Kord Emblem"]. Not all cards are affected: Wear // Tear is fine.

ZeldaZach:
Fix #91 
Fix #90 

This fixes the split card issues with weird values in the "names" field. The issue with Meld card is also somewhat addressed here, given the following structure:
```
Card A
Combined Card
Card B
```

This format seems to hold consistent with all EMN meld cards that I looked at. You can view the compiled file at [EMN.json.txt](https://github.com/mtgjson/mtgjson4/files/2531012/EMN.json.txt)

Two other files I was testing with: [AKH.json.txt](https://github.com/mtgjson/mtgjson4/files/2531018/AKH.json.txt) and [C18.json.txt](https://github.com/mtgjson/mtgjson4/files/2531019/C18.json.txt)

### DIFF
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -370,11 +370,10 @@ def build_mtgjson_card(  # pylint: disable=too-many-branches
     # Handle meld issues
     if "all_parts" in sf_card:
         mtgjson_card["names"] = []
-        for a_part in sf_card["all_parts"]:
-            if "//" in a_part.get("name"):
-                mtgjson_card["names"] = a_part.get("name").split(" // ")
-                break
-            mtgjson_card["names"].append(a_part.get("name"))
+        for card_part in sf_card["all_parts"]:
+            mtgjson_card["names"].extend(
+                name.strip() for name in card_part.get("name", "").split("//")
+            )
 
     # Characteristics that we cannot get from Scryfall
     # Characteristics we have to do further API calls for
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -370,6 +370,9 @@ def build_mtgjson_card(  # pylint: disable=too-many-branches
     if "all_parts" in sf_card:
         mtgjson_card["names"] = []
         for a_part in sf_card["all_parts"]:
+            if "//" in a_part.get("name"):
+                mtgjson_card["names"] = a_part.get("name").split(" // ")
+                break
             mtgjson_card["names"].append(a_part.get("name"))
 
     # Characteristics that we cannot get from Scryfall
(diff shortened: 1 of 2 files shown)


ZeldaZach:
cc: @KingSupernova31

mtgjson4/compile_mtg.py
```diff
@@ -370,6 +371,9 @@ def build_mtgjson_card(  # pylint: disable=too-many-branches
     if "all_parts" in sf_card:
         mtgjson_card["names"] = []
         for a_part in sf_card["all_parts"]:
+            if "//" in a_part.get("name"):
+                mtgjson_card["names"] = a_part.get("name").split(" // ")
+                break
```
gwax, line 376: Can we drop the `if` condition? A split on a string that isn't contained in the string will return a list of a single element.

Can we strip the parts instead of splitting with slashes and whitespace?

For example, we could collapse this to:

```python
if "all_parts" in sf_card:
    mtgjson_card["names"] = []
    for card_part in sf_card["all_parts"]:
        mtgjson_card["names"].extend(name.strip() for name in card_part.get("name", "").split("//"))
```

gwax (approved):
SGTM, with a minor suggestion

KingSupernova31:
The old format in v3 on meld cards was

Card A
Card B
Combined Card

and I think that's the one that people would intuitively expect. However as long as it's consistent, it doesn't really matter.

ZeldaZach:
@KingSupernova31 The system from scryfall returns it in the way i described above, so we will make a documentation notice about the change in format (Since v3->v4 isn't necessarily compatible)

mtgjson4/compile_mtg.py
ZeldaZach: So the answer for right now is no, because if we find a // we stop there (no split cards have a 3rd component).

ZeldaZach:
cc: @ancestral can you add documentation about meld cards and the layout in the "names" field?

PR #107, opened 2018-10-31
Re-implement booster field

### LINKED ISSUE #46: Reimplement Booster Field
The Booster field from V3 contained how cards could be obtained via packs. We removed it in V4, temporarily, to reevaluate our options and figure out the best way to implement.

This will be our discussion board on this topic. 

Cite: https://github.com/mtgjson/mtgjson/issues/521

ZeldaZach:
Fix #46 

This is a cut and paste from V3. I created a 'database' of the booster fields from the v3 files and then turned it into a dictionary that can be referenced by building sets. This will have to be manually maintained for the time being. 

An example is [GRN.json.txt](https://github.com/mtgjson/mtgjson4/files/2532287/GRN.json.txt)


ZeldaZach:
Question: Should we call this field `booster_v3` or something similar, so if we ever choose to revisit this field to improve upon it, we don't break people who depend upon our prior v4 builds? We can always add a `booster` field with our modifications later, and when that's complete, deprecate `booster_v3`. Looking for feedback on this idea

ancestral:
How granular and specific do we want to go with this set meta data? In GRN, each booster has not just a land, but a Gate. In other boosters, like the Khans sets, there is guaranteed to be a dual land. Other sets might have a basic land.

So should these values be defined, or just informational and descriptive?

ZeldaZach:
For now, this is just the reimplementation of the v3 field, as it seems to be a deal breaker for many. I want to come up with our own solution for a better booster field, but we'd have to sit on it for a bit to come up with a strong solution. What do you think of the field rename for legacy purposes?

PR #108, opened 2018-10-31
add a way to distinguish non-unique named cards (mainly un-sets >_>)

### LINKED ISSUE #94: Card entires with the same name are collapsed in allCards.json
Some card entries in the individual set files have the exact same key. ("Everythingamajig", "Ineffable Blessing", etc.) When creating allCards.json, all of these entries are improperly collapsed, leading to a single entry that only has the correct text of one of the variants. (Except it will have a "names" field containing an array of its own name several times.)

ZeldaZach:
Fix #94 

This will convert all cards that have multiple cards with the same name (like `Very Cryptic Command` and `Everythingamajig`) to
```
Very Cryptic Command (a)
Very Cryptic Command (b)
Very Cryptic Command (c)
...
```

There's prob a better way to do what I'm doing, but this seems to be functional. [AllCards.json.txt](https://github.com/mtgjson/mtgjson4/files/2536240/AllCards.json.txt)


KingSupernova31:
This seems to have fixed the collapsing of the entries, but they still have a "names" array that's just their own name as many times as it appeared in the set file.

PR #111, opened 2018-11-05
Add letters to appending numbers for SPLIT-type cards, excluding meld

### LINKED ISSUE #110: Identical card numbers for transform cards
Hi,

I was expecting card numbers to be different for front and back of transform cards (like in v3), but found
- Search for Azcanta / Azcanta, the Sunken Ruin, both number 74 (or 74s in different set)
- Arguel's Blood Fast / Temple of Aclazotz, both 90 (or 90s in a different set)

in v3 this is 74a / 74b and 90a / 90b

Trying to find if a card is front or backside, any advice how this should be done in v4?

thanks!

ZeldaZach:
Fix #110 

~~This adds a letter to split cards' number field to help distinguish them better. Follows V3 formats. Will NOT append letters to meld cards.~~

~~Example: [DGM.json.txt](https://github.com/mtgjson/mtgjson4/files/2550406/DGM.json.txt)~~

This will add a "side" key with value "a", "b", ... for the side/face of the split card. Does NOT affect number

### DIFF
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -347,8 +347,14 @@ def build_mtgjson_card(  # pylint: disable=too-many-branches
     mtgjson_card["isOnlineOnly"] = sf_card.get("digital")  # bool
     mtgjson_card["isOversized"] = sf_card.get("oversized")  # bool
     mtgjson_card["layout"] = sf_card.get("layout")  # str
-    mtgjson_card["number"] = sf_card.get("collector_number")  # str
     mtgjson_card["isReserved"] = sf_card.get("reserved")  # bool
+
+    # Append a lower-case letter to the number if it's a "split" type card
+    # Will only work for two faced cards (not meld, as they don't need this)
+    mtgjson_num = sf_card.get("collector_number")
+    if "names" in mtgjson_card and len(mtgjson_card["names"]) == 2:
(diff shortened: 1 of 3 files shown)


tooomm:
Probably a stupid question, but why not meld?
How was meld done in v3?

ZeldaZach:
Meld aren’t split cards. They’re two individual cards that combine to make something.

tooomm:
Hmm, they reference each other and their front/back sides.
Very similar to other double faced "split" cards, no?

I can't tell if it would be useful to have it for meld to. That's a different question.

ZeldaZach:
We're only talking about card numbers here. Wear // Tear is _one card_ but we treat it as two, as such we call one side "103A" and the other "103B". With meld, they are completely independent cards that _happen to merge_ together. As such, they have unique numbers.

tooomm:
Sure, both meld parts are individual cards with their own number. Right.
But what about the big combined card?
In v3 it used to have a `a-b` relation in the number to one of the "front" cards.

ZeldaZach:
@tooomm Discontinued then as that doesn't make much sense

ancestral:
My concern is that changing the number by adding a letter is not consistent with Wizard’s current numbering scheme. (The numbers listed at the bottom don’t have letters anymore.)

I would rather favor a different key or property to distinguish them as it doesn’t disturb the integrity of the Gatherer data.

Also, I think the broader question is face vs. card — which is it that MTGJSON wants to catalog with?

PR #129, opened 2018-11-22
Fix Names and AllCards

### LINKED ISSUE #116: 8000 cards are duplicated in the file
Around 8000 cards incorrectly have two entries in AllCards.JSON, one labeled with (a) and one with (b). Burning Wish as one example. This affects regular cards along with split cards, double-faced cards, etc.

### LINKED ISSUE #119: Some cards have a nonsensical "names" array
Helm of Kaldra, Hero of Bladehold, and Herald of Anafenza for example.

ZeldaZach:
Fix #119 
Fix #116 

Fixes two PRs in one issue. Will no longer create erroneous entries into the AllCards output, as well as address the "names" still having tokens.

### DIFF
--- a/mtgjson4/__main__.py
+++ b/mtgjson4/__main__.py
@@ -176,9 +176,10 @@ def create_all_cards(files_to_ignore: List[str]) -> Dict[str, Any]:
             duplicate_cards: Dict[str, int] = {}
 
             for card in file_content["cards"]:
+                # Only if a card is duplicated in a set will it get the (a), (b) appended
                 if (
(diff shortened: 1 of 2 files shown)


ZeldaZach:
Future Fix (as this will still not work for 60 tokens): within `all_parts`, it currently has object, id, name, and url. Can you add a new field, `type(string)` that will denote whether it's a card (like a face of a card) or a token?" SF seems interested in the idea of a new field.

Will re-address in the future

tooomm:
Shouldn't this see its own issue then?
I feel like this will be lost here

PR #130, opened 2018-11-22
KeyWords fancy fix

### LINKED ISSUE #120: Keywords.json uses incorrect apostrophe character
The Magic comp rules and therefore the Keywords.json file uses "’" as its apostrophe character (unicode: single right quote, 0x2019). However the oracle text files use "'" (unicode: apostrophe, 0x27) as their apostrophe character. This leads to false negatives when doing any sort of comparison between the two.

Keywords.json should convert all single right quotes to apostrophes.

ZeldaZach:
Fix #120 

This will replace any non-ascii with the ascii equivalent.

### DIFF
--- a/mtgjson4/provider/wizards.py
+++ b/mtgjson4/provider/wizards.py
@@ -7,6 +7,8 @@ from typing import Dict, List
 
 from mtgjson4 import util
 
+import unidecode
+
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


ZeldaZach:
This _only_ does replacements in the Keywords file, no where else in the program

KingSupernova31:
Yeah, I wasn't thinking. :P

PR #143, opened 2018-11-26
Address faceCMC

### LINKED ISSUE #142: Reconsidering CMC for Split Cards
After talking with several parties, I am rethinking how we should handle split face cards CMCs.

On one hand, I agree that a card should list its CMC based on the mana cost. On the other hand, we should help convey the information of the split card by listening the combined CMC too.

My suggestion is we have two fields:

- `convertedManaCost` = the cost of the entire object. An object is a singular card. For `Far // Away` this field would be 5.
- `faceConvertedManaCost` = the cost of this portion. For `Far // Away` the field would be 2 and 3, respectively.  This field would only be available for split cards (and possibly rotation ones?).

Looking for some feedback, as this would be a 4.1.1 change if I don't get further inputs

ZeldaZach:
Fix #142 

Uses the SF built in CMC to determine the mana cost of the cards (so it will be as expected).

Add faceCMC which will be based on the face of the card. 

- Far // Away will report 2 and 3 for face.
- Delver will report 1 and 0 for the faces.
- Budoka Gardener and Dokai will be 2 and 0 for the faces.

[DGM.json.txt](https://github.com/mtgjson/mtgjson4/files/2617652/DGM.json.txt)
[V17.json.txt](https://github.com/mtgjson/mtgjson4/files/2617653/V17.json.txt)
[C18.json.txt](https://github.com/mtgjson/mtgjson4/files/2617676/C18.json.txt)

### DIFF
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -326,8 +326,11 @@ def build_mtgjson_card(
             mtgjson_card["colors"] = get_card_colors(
                 sf_card["mana_cost"].split(" // ")[sf_card_face]
             )
-            mtgjson_card["convertedManaCost"] = get_cmc(
+            mtgjson_card["faceConvertedManaCost"] = get_cmc(
                 sf_card["mana_cost"].split(" // ")[sf_card_face]
+            )
+            mtgjson_card["convertedManaCost"] = get_cmc(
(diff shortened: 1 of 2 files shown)


tooomm:
This needs documentation.

CC: @ancestral

PR #147, opened 2018-11-30
Re-implement variations field

### LINKED ISSUE #145: Not happy with the (a) (b)
https://github.com/Cockatrice/Cockatrice/pull/3458#issuecomment-442996478

Gonna re-think this. Might go back to the variations field for non un-sets instead. Seems like the right approach

ZeldaZach:
Fix #145 

Change the (a), (b), ... implementation for non un-sets to be a variations field instead since they're the same card in every other printing, even if multiple times in a set.

Attached is the un-set that's affected (only one) and a set that exemplifies the variations field.
[UST.json.txt](https://github.com/mtgjson/mtgjson4/files/2632157/UST.json.txt)
[WC99.json.txt](https://github.com/mtgjson/mtgjson4/files/2632158/WC99.json.txt)

### DIFF
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -84,36 +84,54 @@ def build_output_file(sf_cards: List[Dict[str, Any]], set_code: str) -> Dict[str
 def uniquify_duplicates_in_set(cards: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
     """
     For cards with multiple printings in a set, we need to identify
-    them against each other. We will add (a), (b), ... to the end
+    them against each other.
+    For silver border sets, we will add (a), (b), ... to the end
     of the card name to do so.
+    For non-silver bordered sets, we will create a "variations"
+    field will be created that has UUID of repeat cards
     :param cards: Cards to check and update for repeats
     :return: updated cards list
     """
-    unique_list = []
-    duplicate_cards: Dict[str, int] = {}
-    for card in cards:
-        # Only if a card is duplicated in a set will it get the (a), (b) appended
-        total_same_name_cards = len(
-            [item for item in cards if item["name"] == card["name"]]
-        )
+    if cards[0].get("borderColor", None) == "silver":
+        unique_list = []
+        duplicate_cards: Dict[str, int] = {}
+        for card in cards:
+            # Only if a card is duplicated in a set will it get the (a), (b) appended
+            total_same_name_cards = len(
+                [item for item in cards if item["name"] == card["name"]]
+            )
 
-        if (card["name"] not in mtgjson4.BASIC_LANDS) and (
-            card["name"] in duplicate_cards or total_same_name_cards > 1
-        ):
-            if card["name"] in duplicate_cards:
-                duplicate_cards[card["name"]] += 1
+            if (card["name"] not in mtgjson4.BASIC_LANDS) and (
+                card["name"] in duplicate_cards or total_same_name_cards > 1
+            ):
+                if card["name"] in duplicate_cards:
+                    duplicate_cards[card["name"]] += 1
+                else:
+                    duplicate_cards[card["name"]] = ord("a")
+
+                # Update the name of the card, and remove its names field (as it's not correct here)
+                new_card = copy.deepcopy(card)
+                new_card["name"] += " ({0})".format(
+                    chr(duplicate_cards[new_card["name"]])
+                )
+                new_card.pop("names", None)
+                unique_list.append(new_card)
             else:
-                duplicate_cards[card["name"]] = ord("a")
-
-            # Update the name of the card, and remove its names field (as it's not correct here)
-            new_card = copy.deepcopy(card)
-            new_card["name"] += " ({0})".format(chr(duplicate_cards[new_card["name"]]))
-            new_card.pop("names", None)
-            unique_list.append(new_card)
-        else:
-            unique_list.append(card)
-
-    return unique_list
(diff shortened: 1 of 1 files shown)


mtgjson4/compile_mtg.py
```diff
@@ -84,36 +84,54 @@ def build_output_file(sf_cards: List[Dict[str, Any]], set_code: str) -> Dict[str
 def uniquify_duplicates_in_set(cards: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
     """
     For cards with multiple printings in a set, we need to identify
-    them against each other. We will add (a), (b), ... to the end
+    them against each other.
+    For silver border sets, we will add (a), (b), ... to the end
```
Daenyth, line 88: Why do this for silver sets?

mtgjson4/compile_mtg.py
```diff
@@ -84,36 +84,54 @@ def build_output_file(sf_cards: List[Dict[str, Any]], set_code: str) -> Dict[str
...
     :return: updated cards list
     """
-    unique_list = []
-    duplicate_cards: Dict[str, int] = {}
-    for card in cards:
-        # Only if a card is duplicated in a set will it get the (a), (b) appended
-        total_same_name_cards = len(
-            [item for item in cards if item["name"] == card["name"]]
-        )
+    if cards[0].get("borderColor", None) == "silver":
```
Daenyth, line 95: This assumes the input is given homogenous border color - is it conceivable that it wouldn't be?

mtgjson4/compile_mtg.py
```diff
@@ -84,36 +84,54 @@ def build_output_file(sf_cards: List[Dict[str, Any]], set_code: str) -> Dict[str
...
+        unique_list = []
+        duplicate_cards: Dict[str, int] = {}
+        for card in cards:
+            # Only if a card is duplicated in a set will it get the (a), (b) appended
+            total_same_name_cards = len(
```
Daenyth: There's little reason to construct a whole list eagerly and then discard the values just to count the length - you can do `sum(1 for item in cards if ..)`

mtgjson4/compile_mtg.py
```diff
@@ -84,36 +84,54 @@ def build_output_file(sf_cards: List[Dict[str, Any]], set_code: str) -> Dict[str
...
-                duplicate_cards[card["name"]] += 1
+            if (card["name"] not in mtgjson4.BASIC_LANDS) and (
+                card["name"] in duplicate_cards or total_same_name_cards > 1
+            ):
+                if card["name"] in duplicate_cards:
+                    duplicate_cards[card["name"]] += 1
+                else:
+                    duplicate_cards[card["name"]] = ord("a")
+
+                # Update the name of the card, and remove its names field (as it's not correct here)
```
Daenyth, line 113: I find all the conditional logic in this function hard to follow. I'd recommend adding a unit test - the fact that this behavior change doesn't trigger a test failure means that the coverage could use improvement here.

mtgjson4/compile_mtg.py
```diff
@@ -84,36 +84,54 @@ def build_output_file(sf_cards: List[Dict[str, Any]], set_code: str) -> Dict[str
...
+    else:
+        for card in cards:
+            repeats_in_set = [
+                item
+                for item in cards
+                if item["name"] == card["name"] and item["uuid"] != card["uuid"]
+            ]
+            variations = []
+            for repeat in repeats_in_set:
+                variations.append(repeat["uuid"])
```
Daenyth: `variations = [r["uuid"] for r in repeats_in_set]`

Daenyth (commented):
TBH I can't follow the heavily stateful logic in here - it's really confusing for me (too much FP lately :P)

I'd recommend adding in a unit test, something like:

```python
card = {...}
cards_in_set = [card, card, other_card]
unique_cards = uniquify_duplicates_in_set(cards_in_set)
assert len(unique_cards) == 2
```

mtgjson4/compile_mtg.py
ZeldaZach: Because Everythingamagigs, etc from Unstable are all _different_ cards compared to different printings

mtgjson4/compile_mtg.py
ZeldaZach: You are correct that there could be non-silver, but I haven't seen a set where card 1 is non-silver. If you can think of a better way to check `(if any card has borderColor == silver, do this)` lmk

mtgjson4/compile_mtg.py
ZeldaZach: Didn't know this was a thing

mtgjson4/compile_mtg.py
ZeldaZach: At some point I will get back into adding unit tests, won't be in this PR tho :(

mtgjson4/compile_mtg.py
Daenyth: Reasonable enough. As long as you can guarantee that all cards passed in belong to the same set, that's passable.

mtgjson4/compile_mtg.py
Daenyth: The one consistent thing in magic is that nothing is consistent

ZeldaZach: In black border magic, two cards cannot have the same name if they don't have the same rules text. This was thrown out for Unstable, so I added a blanket clause for all un-sets

PR #150, opened 2018-12-02
Move rogue tokens from cards -> tokens array

### LINKED ISSUE #149: Cards without multiverseId
Hi,

I've recently built all the sets and I have discovered that some cards are without a multiverseId.

Sets affected:
- Anthologies - all cards are without it
- DeckMasters - all cards are without it
- Duel Decks: Elves vs Inventors - only the card `Darksteel Plate`
- Duel Decks: Zendikar vs Eldrazi - The token`plant` is in the array of cards
- Duels of the Planeswalkers - all cards are without it
- Explorers of Ixalan - The token `Saproling` is in the array of cards
- From the Vault: Lore - only the card `Marit Lage`
- Tenth edition - all cards are without it
- Ultimate Masters - all cards are without it
- Unhinged - all cards version (b), eg. `Cardpecker (a)` has it, `Cardpecker (b)` doesn't

I've checked all the pre-built json as well at version `4.1.2`. 
Let me know if you need any more info!

Thanks :)

ZeldaZach:
Fix #149 

Sometimes tokens slip through and join the normal sets. We will break them up and put them (correctly) into the tokens array.

[DDP.json.txt](https://github.com/mtgjson/mtgjson4/files/2637362/DDP.json.txt)
[E02.json.txt](https://github.com/mtgjson/mtgjson4/files/2637363/E02.json.txt)
[V16.json.txt](https://github.com/mtgjson/mtgjson4/files/2637364/V16.json.txt)

### DIFF
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -71,7 +71,7 @@ def build_output_file(sf_cards: List[Dict[str, Any]], set_code: str) -> Dict[str
 
     # Address duplicate printings in a set
     card_holder = uniquify_duplicates_in_set(card_holder)
-    
+
     # Move bogus tokens out
     card_holder, added_tokens = transpose_tokens(card_holder)
 
@@ -92,7 +92,7 @@ def build_output_file(sf_cards: List[Dict[str, Any]], set_code: str) -> Dict[str
 
     return output_file
 
- 
+
 def transpose_tokens(
     cards: List[Dict[str, Any]]
 ) -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]:
@@ -115,7 +115,7 @@ def transpose_tokens(
 
     return cards, tokens
 
-  
+
 def add_tcgplayer_ids(
     group_id: int, cards: List[Dict[str, Any]]
 ) -> List[Dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


mtgjson4/compile_mtg.py
```diff
@@ -75,12 +77,35 @@ def build_output_file(sf_cards: List[Dict[str, Any]], set_code: str) -> Dict[str
...
+    """
+    # Order matters with these, as if you do cards first
+    # it will shadow the tokens lookup
+    tokens = [
+        scryfall.download(scryfall.SCRYFALL_API_CARD + card["uuid"])
+        for card in cards
+        if card["layout"] == "token"
+    ]
+
+    cards = [card for card in cards if card["layout"] != "token"]
```
ebbit1q, line 114: Wouldn't it be more efficient to just change the old list?
like:
```
indexes = [i for i, card in enumerate(cards) if card == "token"]
tokens = [cards.pop(i) for i in indexes[::-1]]
```

ZeldaZach: The list is no more than 500 elements, so it shouldn't be a big deal. 

I'm also not 100% certain what your iteration entails... looks like you're reverse-iterating the cards array and popping at the found index... I don't believe you can remove from a list while iterating

ebbit1q: Well, I agree this isn't a very nice way to do it, but you can remove items in a loop as long as the indexes still exist (that's why it's in reverse, sort of wrote that quickly)

PR #151, opened 2018-12-02
add version+date badge

tooomm:
See https://github.com/tooomm/mtgjson-python/blob/patch-1/README.md

They will updated from the `version.json` file at the webpage.


ZeldaZach:
Appreciated!!

PR #183, opened 2018-12-18
split duel decks into a/b decks ("duelDeck": "LETTER")

### LINKED ISSUE #75: Add 'deck' or 'subset' field
This can be used to identify which cards are part of which decks (such as 'elf' deck in the 'elves vs goblins' set). The existing 'source' field is used a bit like this for some promos sets like CPK, but it's not quite the same thing, but it probably would work just fine for this.

https://github.com/mtgjson/mtgjson/issues/51

ZeldaZach:
Fix #75 

Split duel decks via a new field: `duelDeck (string)` which contains either "a" or "b" to indicate what deck it's in.

The way this is accomplished (right now) is to find the basic land string, then split it. [Card 1 .. Last Basic in first string of basics] U [Next card .. Card N]

Files:
[DDT.json.txt](https://github.com/mtgjson/mtgjson4/files/2689634/DDT.json.txt)
[DDU.json.txt](https://github.com/mtgjson/mtgjson4/files/2689635/DDU.json.txt)

### DIFF
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -101,7 +101,7 @@ def build_output_file(
     # Add Variations to each entry
     add_variations_field(output_file["cards"])
 
-    if "DD" == set_code[:2]:
+    if set_code[:2] == "DD":
         mark_duel_decks(output_file["cards"])
 
     return output_file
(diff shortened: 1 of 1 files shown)


cheyner:
👍

PR #228, opened 2019-01-21
fixed issue where TCGPlayer property queries failed for certain cards

kellenwhitaker:
Fixed issue where TCGPlayer property queries for basic lands, GRN/RNA Guildgates, split cards, and cards having diacritical marks in their names failed

### DIFF
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -206,6 +206,7 @@ def add_tcgplayer_fields(
     For each card in the set, we will find its tcgplayer ID
     and add it to the card if found
     :param group_id: group to search for the cards
+    :param set_code: the set code associated with the cards
     :param cards: Cards list to add information to
     :return: Cards list with new information added
     """
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -85,13 +85,7 @@ def build_output_file(
     # Move bogus tokens out
     card_holder, added_tokens = transpose_tokens(card_holder)
 
-    # Add TCGPlayer information
-    if "tcgplayer_id" in set_config:
-        output_file["tcgplayerGroupId"] = set_config.get("tcgplayer_id")
-        if not skip_tcgplayer:
-            card_holder = add_tcgplayer_fields(
-                output_file["tcgplayerGroupId"], card_holder
-            )
+    # Add TCGPlayer information    if "tcgplayer_id" in set_config:        output_file["tcgplayerGroupId"] = set_config.get("tcgplayer_id")        if not skip_tcgplayer:            card_holder = add_tcgplayer_fields(                output_file["tcgplayerGroupId"], output_file["code"], card_holder            )
 
     # Set sizes; BASE SET SIZE WILL BE UPDATED BELOW
     output_file["totalSetSize"] = len(sf_cards)
@@ -199,28 +193,7 @@ def transpose_tokens(
     return cards, tokens
 
 
-def add_tcgplayer_fields(
-    group_id: int, cards: List[Dict[str, Any]]
-) -> List[Dict[str, Any]]:
-    """
-    For each card in the set, we will find its tcgplayer ID
-    and add it to the card if found
-    :param group_id: group to search for the cards
-    :param cards: Cards list to add information to
-    :return: Cards list with new information added
-    """
-    tcg_card_objs = tcgplayer.get_group_id_cards(group_id)
-
-    for card in cards:
-        prod_id = tcgplayer.get_card_property(card["name"], tcg_card_objs, "productId")
-        prod_url = tcgplayer.get_card_property(card["name"], tcg_card_objs, "url")
-
-        if prod_id and prod_url:
-            card["tcgplayerProductId"] = prod_id
-            card["tcgplayerPurchaseUrl"] = tcgplayer.log_redirection_url(
-                card["tcgplayerProductId"], prod_url
-            )
-
+def add_tcgplayer_fields(    group_id: int, set_code: str, cards: List[Dict[str, Any]]) -> List[Dict[str, Any]]:    """    For each card in the set, we will find its tcgplayer ID    and add it to the card if found    :param group_id: group to search for the cards    :param cards: Cards list to add information to    :return: Cards list with new information added    """    tcg_card_objs = tcgplayer.get_group_id_cards(group_id)    for card in cards:        query_name = tcgplayer.get_card_query_name(card, set_code)        prod_id = tcgplayer.get_card_property(query_name, tcg_card_objs, "productId")        prod_url = tcgplayer.get_card_property(query_name, tcg_card_objs, "url")        if prod_id and prod_url:            card["tcgplayerProductId"] = prod_id            card["tcgplayerPurchaseUrl"] = tcgplayer.log_redirection_url(                card["tcgplayerProductId"], prod_url            )
     return cards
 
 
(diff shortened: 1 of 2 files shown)


mtgjson4/compile_mtg.py
```diff
@@ -200,7 +200,7 @@ def transpose_tokens(
 
 
 def add_tcgplayer_fields(
-    group_id: int, cards: List[Dict[str, Any]]
+    group_id: int, set_code: str, cards: List[Dict[str, Any]]
```
ZeldaZach, line 203: Function docs must be updated

mtgjson4/provider/tcgplayer.py
```diff
@@ -145,23 +146,43 @@ def get_group_id_cards(group_id: int) -> List[Dict[str, Any]]:
...
+) -> str:
+    """
+    Returns the TCGPlayer-compatible query name. Removes diacritical marks. 
+    Appends collector number to basic lands, and to GRN and RNA Guildgates.
+    :param card: MTGJSON card to find query name for
+    :param set_code: The code of the set the MTGJSON card belongs to
+    :return: TCGPlayer-compatible query name
+    """
+    unidecoded_name = unidecode.unidecode(card["name"])
+    if (unidecoded_name == "Plains" or unidecoded_name == "Island" or unidecoded_name == "Swamp"
```
ZeldaZach: `if (x in [a,b,c,...])` would be better

mtgjson4/provider/tcgplayer.py
```diff
@@ -145,23 +146,43 @@ def get_group_id_cards(group_id: int) -> List[Dict[str, Any]]:
...
-    :param card_name: Card name to find in the list
+    :param query_name: TCGPlayer-compatible card name to find in the list
     :param card_list: List of TCGPlayer card objects
     :param card_field: Field to pull from TCGPlayer card object
     :return: Value of field
     """
     for card in card_list:
-        if card_name.lower() == card["name"].lower():
+        # will try to match against Part A for split cards (if applicable)
+        part_a_name = card["name"].split(" //")[0]
```
ZeldaZach: How do I handle splits elsewhere? I'm pretty sure I use a trim as it might _not_ be `_//` but just a `//`

mtgjson4/provider/tcgplayer.py
```diff
@@ -145,23 +146,43 @@ def get_group_id_cards(group_id: int) -> List[Dict[str, Any]]:
...
+    :param query_name: TCGPlayer-compatible card name to find in the list
     :param card_list: List of TCGPlayer card objects
     :param card_field: Field to pull from TCGPlayer card object
     :return: Value of field
     """
     for card in card_list:
-        if card_name.lower() == card["name"].lower():
+        # will try to match against Part A for split cards (if applicable)
+        part_a_name = card["name"].split(" //")[0] 
+        if query_name.lower() == part_a_name.lower():
```
ZeldaZach, line 181: I think there's a way to just compare w/o case sensitivity

ZeldaZach (commented):
Appreciate the PR! Always great to have new people helping around here :)

I gave an initial overview, without testing it. Please try to fix those up and then i'll give this a shot!

mtgjson4/provider/tcgplayer.py
kellenwhitaker: `_//` is the TCGPlayer "standard." Point taken, though, there is no such thing as a standard unless it's an explicit one. Will change this.

mtgjson4/provider/tcgplayer.py
kellenwhitaker: I don't know what you mean here. The code is the same as before. Just different vars.

mtgjson4/provider/tcgplayer.py
kellenwhitaker: Changed.

mtgjson4/compile_mtg.py
kellenwhitaker: Updated. Triple-checked all function docs.

mtgjson4/provider/tcgplayer.py
```diff
@@ -146,22 +147,41 @@ def get_group_id_cards(group_id: int) -> List[Dict[str, Any]]:
...
+    Returns the TCGPlayer-compatible query name. Removes diacritical marks.
+    Appends collector number to basic lands, and to GRN and RNA Guildgates.
+    :param card: MTGJSON card to find query name for
+    :param set_code: The code of the set the MTGJSON card belongs to
+    :return: TCGPlayer-compatible query name
+    """
+    unidecoded_name = unidecode.unidecode(card["name"])
+    if unidecoded_name in ["Plains", "Island", "Swamp", "Mountain", "Forest"]:
+        return "{} ({})".format(unidecoded_name, card["number"])
+    if set_code in ["GRN", "RNA"] and "Guildgate" in unidecoded_name:
```
ZeldaZach, line 161: Is this really only a problem with these two sets? This function feels very hacky imo. Would it be better to decode all entries and try to fit them into TCGPlayer?

kellenwhitaker: It is a little hacky. Those are the only two sets where the Guildgates have variations. So you're maybe proposing something like trying both the original name and the name with collector number appended? That would encompass both the basic land and GRN/RNA Guildgate cases. It would probably help future proof as well. Then what about original vs. unidecoded name?

kellenwhitaker: Checking the data, there are currently no cases where a name with accents has its collector number appended in TCGPlayer. To help future proof, though, the number of checks could be expanded to 4: original name, appended name, unidecoded name, appended unidecoded name. This would eliminate the above function entirely. What do you think?

mtgjson4/provider/tcgplayer.py
```diff
@@ -146,22 +147,41 @@ def get_group_id_cards(group_id: int) -> List[Dict[str, Any]]:
...
     for card in card_list:
-        if card_name.lower() == card["name"].lower():
+        # will try to match against Part A for split cards (if applicable)
+        part_a_name = card["name"].split("//")[0].strip()
+        if query_name.lower() == part_a_name.lower():
             return card.get(card_field, None)
 
-    LOGGER.warning("Unable to find card {} in TCGPlayer card list".format(card_name))
+    LOGGER.warning("Unable to find card {} in TCGPlayer card list".format(query_name))
     return None
```
kellenwhitaker, line 187: So, eliminating `get_card_query_name()`, the new `get_card_property()` would look something like this:
```
def get_card_property(
    mtgjson_card: Dict[str, Any], card_list: List[Dict[str, Any]], card_field: str
) -> Any:
    """
    Go through the passed in card object list to find the matching
    card from the set and get its attribute.
    :param mtgjson_card: Card to find in the list
    :param card_list: List of TCGPlayer card objects
    :param card_field: Field to pull from TCGPlayer card object
    :return: Value of field
    """
    for card in card_list:
        # will try to match against Part A for split cards (if applicable)
        part_a_name_lowered = card["name"].split("//")[0].strip().lower()
        if mtgjson_card["name"].lower() == part_a_name_lowered:
            return card.get(card_field, None)
        if "{} ({})".format(mtgjson_card["name"], mtgjson_card["number"]).lower() == part_a_name_lowered:
            return card.get(card_field, None)
        unidecoded_name = unidecode.unidecode(mtgjson_card["name"])
        if unidecoded_name.lower() == part_a_name_lowered:
            return card.get(card_field, None)
        if "{} ({})".format(unidecoded_name, mtgjson_card["number"]).lower() == part_a_name_lowered:
            return card.get(card_field, None)

    LOGGER.warning("Unable to find card {} in TCGPlayer card list".format(mtgjson_card["name"]))
    return None
```

ZeldaZach: Something like this looks better to me, as it's based around split card attributes (which is ok), and not hardcoded sets/etc.

ZeldaZach:
Btw, sorry I didn't mention this earlier, can you possibly look into adding a test to the repo to confirm this works as expected? Been looking to add tests, and this would be a great start :)

kellenwhitaker:
> Btw, sorry I didn't mention this earlier, can you possibly look into adding a test to the repo to confirm this works as expected? Been looking to add tests, and this would be a great start :)

I'm not prepared to add a test atm. I did do a manual check on all official US sets. After filling in a few holes, there are 87 cards remaining (all basic lands), from 7 sets--predominantly older ones. The reason for failure falls into one of two categories: either TCGPlayer appends a description of the card art instead of collector number, or there is a mismatch with the collector number. How do you want to proceed? The easy way is to map (uuid)-->(TCGPlayer name), or something similar. The mappings could be placed in a text file, and this would be much easier and less error prone than trying to map them dynamically.

Heads up: there are more cards (not covered here) which exhibit this same issue, namely cards with variations from older sets. TCGPlayer appends a description of the card art on these as well.

ZeldaZach:
If we can't solve all of them through your code method, we can add a resources file with the mapping. I'd rather try to solve them all via code, but if you can only handle a portion then we can look into alternative methods. Feel free to reach out on Discord to chat more :) (Link's on the website)

kellenwhitaker:
Will go with the resources file. The lion's share is handled strictly through code. Just doesn't seem worth developing a whole new parsing engine for a very limited number of affected cards. That method would end up leaving out cards anyhow, since at times TCGPlayer omits the collector number on the page, and the only way to identify the card is the art itself. Nor is any of this information exposed through the API.

Will code this, do some more testing, and do another commit by the end of the day.

PR #234, opened 2019-01-25
add Scryfall as a first source for TCGPlayer Product IDs

kellenwhitaker:
Adding Scryfall as a first source for TCGPlayer Product IDs improves coverage of those IDs. TCGPlayer should be retained as a source since Scryfall data is incomplete.

### DIFF
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -203,7 +203,7 @@ def add_tcgplayer_fields(
     group_id: int, cards: List[Dict[str, Any]]
 ) -> List[Dict[str, Any]]:
     """
-    For each card in the set, we will find its tcgplayer ID
+    For each card in the set, we will find its product URL
     and add it to the card if found
     :param group_id: group to search for the cards
     :param cards: Cards list to add information to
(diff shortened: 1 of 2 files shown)


ZeldaZach:
So it seems your check was incorrect and didn't do what we expected. I glanced too briefly at it, and it's my mistake. Wish we had better testing. Learned of this error after a full rebuild, so now I need to do some creative stuff to fix it up

kellenwhitaker:
Thanks. I was pretty meticulous in my checks (all official US sets). Promos were not tested. Out of > 30,000 cards tested, only a few dozen unrelated cards had missing product IDs. Note this would still leave out a lot of product URLs (another reason I started a new PR instead of going with the old one). I see you added a .split() to the input name. This was never an issue with the sets I tested. Would be curious to know, if you recall, from which sets you found failures.

Yes, we could always use better testing. Along those lines, I'll be putting in a feature request to see if we can speed up compilation.

ZeldaZach:
The issue was it your for loop check would match the first entry every time, so the URLs would all be the same per set. I fixed the check to be similar to what I did before.

It failed with _every_ set, and I'm shocked I missed it.

kellenwhitaker:
Wow. I just checked it out. That's ugly. I was pretty worn out having to deal with Visual Studio's GH extension, but that's something that should not have been missed in any event.

PR #244, opened 2019-02-01
Fix #241: foreignData with data of other cards

### LINKED ISSUE #241: Foreign names generation does not check card 'number' and duplicates entries
Hi, first of all thank for your work with this project. 
While porting an application to version 4, I noticed a possible bug in the generation of `foreignData`.

#### Description of Bug:
In sets where there are cards with different `number`, for example `Plains` cards in `C18` set, the `foreignData` field of each card contains all the `foreignData` of cards with the same `multiverseId` and different `number`.

So for example, generating the `C18` set, if you see the card with data
```json
            "multiverseId": 451248,
            "name": "Plains",
            "number": "293",
```

you can see its `foreignData` field contains multiple entries for each language, for example `Italian`:

```json
                {
                    "language": "Italian",
                    "multiverseId": 452242,
                    "name": "Pianura",
                    "type": "Terra Base — Pianura"
                },
                [...]
                {
                    "language": "Italian",
                    "multiverseId": 452241,
                    "name": "Pianura",
                    "type": "Terra Base — Pianura"
                },
                [...]
                {
                    "language": "Italian",
                    "multiverseId": 452240,
                    "name": "Pianura",
                    "type": "Terra Base — Pianura"
                },
```

This can be fixed by passing `card['number']` to `parse_foreign` function inside `provider/scryfall.py`. 
With this information, the check in line 212 could go from:
```python
if set_name != foreign_card["set"] or foreign_card["lang"] == "en":
    continue
```
to
```python
if (set_name != foreign_card["set"]
    or card['number'] != foreign_card['collector_number']
    or foreign_card["lang"] == "en"):
    continue
```

If you confirm that this is a bug, I can open a PR with the fix.

edit: I just noticed there were some new commits today, but apparently the problem is still present

Psykopear:
PR that fixes #241 
I also added a test to avoid regressions and added pytest and pytest-cov as deps to the tox.ini to make tests run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,8 @@ ___
 | `--skip-sets SET [SET ...]` | Prevents set code arguments from being built, even if passed in via `-a` or `-s`.                                                               |
 | `--skip-tcgplayer`          | If you don't have a TCGPlayer API key, you can disable building of TCGPlayer components.                                                        |
 
-> &nbsp;  
-> **Newcomer Note**: Omitting either the `-a` and `-s` flag will yield empty outputted data.  
+> &nbsp;
+> **Newcomer Note**: Omitting either the `-a` and `-s` flag will yield empty outputted data.
 > &nbsp;
 
 &nbsp;
@@ -87,6 +87,6 @@ ___
 
 ## **Notes**
 
-> &nbsp;  
-> These are the build directions to compile your own JSON files but If you are looking for pre-compiled JSON files, you can download them at [MTGJSON.com](https://mtgjson.com/).  
+> &nbsp;
+> These are the build directions to compile your own JSON files but If you are looking for pre-compiled JSON files, you can download them at [MTGJSON.com](https://mtgjson.com/).
 > &nbsp;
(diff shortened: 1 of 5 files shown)


Psykopear:
I also added some fixes in the `README.md` file so the CI passes

ZeldaZach:
Can you upload a before/after JSON file so I can compare the outputs? If not, I can just run locally (just a bit late rn)

Psykopear:
So this is the file I generated for C18 set:
https://gist.github.com/Psykopear/68958c45f130013f06a38d654654a5bf

And this is the diff between the fixed file and https://mtgjson.com/json/C18.json :
https://gist.github.com/Psykopear/0101b63f246bd2e7dbc81660fb2ee649

In my file the field `tcgplayerPurchaseUrl` is missing, but I think it's just because I didn't add an API key

ZeldaZach:
Just tested this locally. Seems to remove invalid foreignData entries for basic lands, which is correct and a great fix.

I appreciate you taking the time to contribute to my project. Please, feel free to stick around and chat with us on [Discord](https://discord.gg/74GUQDE) at some point :)

PR #245, opened 2019-02-01
Use (ScryfallId, Name) to identify the card face

NikolayXHD:
When other card providers will be implemented in the future
the cards without scryfallId may need a different set of fields
to be uniquely identified


ZeldaZach:
Much appreciated, @NikolayXHD !

I've made some minor changes and added a unit test to prevent future breakage :)

PR #248, opened 2019-02-05
Ticket 247: Fixing Number and layout

excention:
The card_face object doesn't support the same fields as the card object so I adjusted the code to pull the fields from the card object where not available by the card_face.

The Unstable set was the big offender. The double faced tokens were showing up but not all the data was there compared to a single face token.. So in this case I pulled the data applied to main card object for the card faces, and then I added the layout and side data. 

not sure what other sets have double faced tokens. But the Unstable set was where I saw the issue.


excention:
This is untested pseudo code to make sure to test and verify before merging.

ZeldaZach:
Diff file between master and this branch: [UST.txt](https://github.com/mtgjson/mtgjson/files/2833558/UST.txt)

ZeldaZach:
Looking at the changes, it seems like a strict improvement. My only issue is `uuid` are the same for tokens with two sides... that needs to be addressed. Might not be this PR, but _a_ pr is needed.

excention:
Looks good, the scryfall id is the same but it already was the same. The only thing I see on the card face is an illustration_id not sure if that's what you will ultimately want

PR #336, opened 2019-04-29
Add skip key check for stocks data

ZeldaZach:


skoogiz:
Looks good to me. I think its good to have a common flag for all price sites if you want to build without price data.

ZeldaZach:
Agreed

PR #417, opened 2019-07-25
Token "reverseRelated" bug

omfgitsmark:
The "reverseRelated" field for tokens currently lists the related card's names in a comma delimited format but card names can contain commas which will interfere with parsing the field's data. It should probably be a list of the UUIDs of related cards.

### DIFF
--- a/mtgjson4/compile_mtg.py
+++ b/mtgjson4/compile_mtg.py
@@ -536,8 +536,8 @@ def build_mtgjson_tokens(
         reverse_related: List[str] = []
         if "all_parts" in sf_token:
             for a_part in sf_token["all_parts"]:
-                if a_part.get("name") != token_card.get("name"):
-                    reverse_related.append(a_part.get("name"))
+                if a_part.get("uuid") != token_card.get("uuid"):
+                    reverse_related.append(a_part.get("uuid"))
         token_card.set("reverseRelated", reverse_related)
 
         LOGGER.info(f"Parsed {token_card.get('name')} from {sf_token.get('set')}")


ZeldaZach:
This appears to be a breaking change, so I'll have to hold off until 4.5.0 (which is probably a week or two away). Can you upload some examples on before/after w/ a diff file?

omfgitsmark:
**TL;DR:** I'm dumb, this is not really a "bug", but it's still an improvement and valid change.

So, I'm kinda stupid, I guess my brain forgot I've been making my database with mtgsqlive (which converts arrays to comma-delimited strings) and not my old script (which just stored a JSON object for arrays), so when I saw an example of the "reverseRelated" field and it was stored as a comma-delimited string I automatically figured it was stored that way in the MTGJSON builds. So this is not really a "bug", but it could still cause a problem for some. The "reverseRelated" field should still be changed to a list of the related card's UUIDs instead of their names for easier referencing and preventing any parsing issues in user projects.

If you'd like to see what I mean and you have a compiled SQLite database handy, pull up the "Cat" token from M19 and look at the "reverseRelated" field, it's an absolute mess.

`SELECT reverseRelated FROM tokens WHERE name='Cat' AND setCode='M19' LIMIT 1`

> Ajani, Adversary of Tyrants, Leonin Warleader

ZeldaZach:
This change appears to affect a lot of my upstream users. As such, we'll need to create a tag for change and give enough time before rolling such a change in.

Also, reverseRelated by card name makes a bit more sense as it (theoretically) can link to cards outside of the individual set. Unless I'm wrong on that front...

ZeldaZach:
I'll find a better solution. This is a breaking change i'm not ready for.

PR #441, opened 2019-09-12
Issue/Enhancement-47 More Formats

### LINKED ISSUE #47: AllCards/AllSets for more formats
Continuation of https://github.com/mtgjson/mtgjson/issues/178

Currently, we will build all sets in existence into these two files. Should we look into a limiting factor, or have different downloads? Technically, they're all real magic cards, so I'd like to keep them bundled. They have an annotation that they're not legal (silver boarder) and most people can infer that information.

Discussions welcome!

### LINKED ISSUE #428: Crawling Gamepedia is a serious risk
#### Description of Bug:
MTGJSON crawls https://mtg.gamepedia.com/Modern for a list of sets that are legal.

This is insecure. Anyone can edit that page and insert or delete a set before a rebuild occurs. Or less maliciously, the page could easily get designed, reformatted, or redirected without notice.

We need a safer method. Either a trusted site like Scryfall, or another that has an API or a Read-Only page, or maybe MTGJSON should just build in the logic itself.

Solving this may also help #47 . (Vintage, Legacy, Brawl, Commander and others could have their own files.)

henry739:
This PR fixes #47 (and indirectly fixes #428) and ends up with the following:
## Set-Centric Outputs
- Standard.json
- Modern.json
- Legacy.json
- Vintage.json

## Card-Centric Outputs
- StandardCards.json
- ModernCards.json
- LegacyCards.json
- VintageCards.json
- PauperCards.json

## Other
In addition, the old implementations for retrieving modern sets (from Gamepedia) and standard have been replaced with a more general approach that uses data we already have.


mtgjson4/format.py
```diff
@@ -0,0 +1,50 @@
+"""
+Functions and constants used to help build the format map
+"""
+from typing import List, Dict, Any
+
+FORMATS = (
```
ZeldaZach: Define as set via brackets, since we use it that way later in the code

ZeldaZach: Also, define the type `Set[str]`

mtgjson4/format.py
```diff
@@ -0,0 +1,50 @@
...
+)
+
+
+def build_format_map(all_sets: Dict[str, Any], regular: bool = True) -> Dict[str, List[str]]:
+    """
+    For each set in the specified JSON file, determine its legal sets and return a dictionary mapping set code to
+    a list of legal formats.
+
+    :param all_sets: AllSets content
+    :type all_sets: dict
```
ZeldaZach: I don't expect types in the formats, just params :)

mtgjson4/format.py
```diff
@@ -0,0 +1,50 @@
...
+    :param regular: If this is True, then unusual sets will be excluded.
+    :type regular: bool
+
+    :return: Dictionary of the form { format: [codes] }
+    :rtype: dict
+    """
+    formats: Dict[str, List[str]] = {fmt: [] for fmt in FORMATS}
+
+    for code, data in all_sets.items():
+        if regular and data["type"] not in ["expansion", "core", "draft_innovation", "commander", "masters"]:
```
ZeldaZach: should global var that list, since it could expand

mtgjson4/format.py
```diff
@@ -0,0 +1,50 @@
...
+            continue
+
+        possible_formats = set(FORMATS)
+        cards = data.get("cards")
+
+        for card in cards:
+            card_formats = set(card.get("legalities").keys())
+            possible_formats &= card_formats
```
ZeldaZach: I'd like to avoid bitwise ops if possible. Any other way of going about this?

mtgjson4/outputter.py
```diff
@@ -349,11 +330,23 @@ def create_and_write_compiled_outputs() -> None:
...
 
+    # Compute format map from all_sets
+    format_map = build_format_map(all_sets)
+
     # Standard.json
-    write_to_file(mtgjson4.STANDARD_OUTPUT, create_standard_only_output())
+    write_to_file(mtgjson4.STANDARD_OUTPUT, __handle_compiling_sets(format_map["standard"], "Standard"))
+
+    # Future.json
+    write_to_file(mtgjson4.FUTURE_OUTPUT, __handle_compiling_sets(format_map["future"], "Future"))
```
ZeldaZach: I'd like to re-key Future => "Future Standard"

tests/mtgjson4/test_format.py
```diff
@@ -0,0 +1,68 @@
+import pytest
+
+from mtgjson4.format import build_format_map
+
+
+NULL_OUTPUT = {
```
ZeldaZach: Type is required. Also, if this is the same as the list before, you can simply make one list and refer to it as keys and/or values.

ZeldaZach: Err this is a tester, so not sure if we'd want to do what I said above. Can go either way.

mtgjson4/format.py
```diff
@@ -0,0 +1,50 @@
...
+    :rtype: dict
+    """
+    formats: Dict[str, List[str]] = {fmt: [] for fmt in FORMATS}
+
+    for code, data in all_sets.items():
+        if regular and data["type"] not in ["expansion", "core", "draft_innovation", "commander", "masters"]:
+            continue
+
+        possible_formats = set(FORMATS)
+        cards = data.get("cards")
```
ZeldaZach: don't need a local var, can just put in the for loop

mtgjson4/format.py
```diff
@@ -0,0 +1,50 @@
...
+
+        for card in cards:
+            card_formats = set(card.get("legalities").keys())
```
ZeldaZach, line 36: I honestly don't know what something like this would report back...

ZeldaZach (changes requested):
Good start! Keep up the great work :)

mtgjson4/format.py
henry739: Sure there is: It can be expressed as `possible_formats = possible_formats.intersection(card_formats)`

mtgjson4/format.py
henry739: I'll add some clarifying comments, but the dictionary keys (format) are only present in the legalities dictionary if the value is != 'Not Legal'.

mtgjson4/outputter.py
henry739: Do you want the output to be called "Future Standard"?

henry739:
Should be ready for re-review now @ZeldaZach. Noted that this should also fix an additional issue (#428 .) 

From looking at `outputting.py`, I feel that some refactoring could be beneficial. I didn't want to block the merging of this work, and if agreed to be beneficial, perhaps I could undertake this separately?

omfgitsmark (commented):
Where is `get_modern_sets()` from gamepedia.py now? Or where is it getting the modern legal sets from?

mtgjson4/outputter.py
```diff
@@ -4,12 +4,13 @@
 import contextvars
 import json
 import logging
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Set
 
 import mtgjson4
-from mtgjson4 import util
+from mtgjson4 import util, SUPPORTED_FORMAT_OUTPUTS
```
omfgitsmark: Is this necessary? `mtgjson4.SUPPORTED_FORMAT_OUTPUTS` should work inline.

omfgitsmark:
**BIG PICTURE:**
Do we really need a set-based and card-based file for each format?

It feels like we have a really bad mixture of dicts, lists, tuples, and now sets. 

I really like the organization and changes added to outputter.py

Have you run this successfully? I feel like it might not even compile.

mtgjson4/outputter.py
henry739: A matter of style maybe? Do we generally want these constants to be prefixed with the module name?

henry739:
@omfgitsmark

- The overall set legality is (should!) be inferred from the card data by the utility function in format.py.
- I don't know if there is a use case for the per-set JSON files? It does seem a little redundant. I wonder if any users depend on this structure?
- I agree about the mixture of data structures - and also the sheer number of globals. I would be up for some refactoring work here - but it makes sense to keep it separate from this issue, right? 
- I carried out developer testing by loading AllCards and AllSets from pre-built JSON files, ran them through my methods that generate the per-format files, and looked at the outputs. I have not run this end to end yet.

mtgjson4/__init__.py
```diff
@@ -53,12 +53,34 @@
...
-MODERN_OUTPUT: str = "Modern"
 REFERRAL_DB_OUTPUT: str = "ReferralMap"
 SET_LIST_OUTPUT: str = "SetList"
-STANDARD_OUTPUT: str = "Standard"
 VERSION_OUTPUT: str = "version"
+
+STANDARD_OUTPUT: str = "Standard"
+MODERN_OUTPUT: str = "Modern"
 VINTAGE_OUTPUT: str = "Vintage"
+LEGACY_OUTPUT: str = "Legacy"
```
ZeldaZach, line 63: I do want to make a `globals` class in the near future since this is getting a bit large for our init

mtgjson4/outputter.py
ZeldaZach: We don't have a style doc for such a thing, but I'm fine with mtgjson4.variable (see my earlier comment about GLOBAL class coming soon :tm: )

mtgjson4/outputter.py
```diff
@@ -165,6 +166,27 @@ def create_all_cards(files_to_ignore: List[str]) -> Dict[str, Any]:
...
+
+    :param cards: The card data to be filtered (should be AllCards)
+    :param target_formats: The set of formats to create subsets for.
+
+    :return: Dictionary of dictionaries, mapping format name -> all_cards subset.
+    """
+    subsets: Dict[str, Dict[str, Any]] = {fmt: {} for fmt in target_formats}
+    for name, data in cards.items():
+        for fmt in target_formats:
+            if data["legalities"].get(fmt) in ("Legal", "Restricted"):
```
ZeldaZach, line 184: Good change to do `if var in set(...)` it seems to be much faster than `if var in list(...)`. I should go back and start profiling this project...

ZeldaZach (changes requested):
I've had a chance to test this and there are a few areas of improvement:

- Your method of narrowing down what formats contain what sets works for Standard and Modern, but it looks to be wrong for Pauper, Legacy, Vintage, etc
- Pauper has some issues since cards are legal from Vintage sets, but the method we use to narrow down sets is whole sets at a time (so the uncommons rares etc will knock out the entire set). Same issue as above

- Pauper.json needs work (it's empty)

Maybe I"m overthinking this one?

ZeldaZach:
I'm happy with this. A good first run :)

PR #459, opened 2019-10-13
Pricing split to separate file

ZeldaZach:
This is the start of file splitting. We will update smaller chunks of MTGJSON on a daily basis and provide those as separate download options.

The main MTGJSON dump will be stripped of most price data. We will continue to server the same key/values as expected, but will only have up to one entry per field (i.e. the latest price on mtgo and in paper, foil/not-foil).

The new `Prices.json` file will be updated daily and is index-able by MTGJSON uuids. That will return a prices array that can be pulled from to enhance an application, if necessary. This cuts down on the download size for most people who may not care about pricing, and gives enhanced options for us to rebuild the system in a cost-effective manner.

Tagging @omfgitsmark for handling SQLite/MySQL changes, as I'm not sure of the logistics on that front currently.

Tagging @staghouse for handling of website related changes.

### DIFF
--- a/mtgjson4/__init__.py
+++ b/mtgjson4/__init__.py
@@ -54,6 +54,7 @@ CARD_TYPES_OUTPUT: str = "CardTypes"
 COMPILED_LIST_OUTPUT: str = "CompiledList"
 DECK_LISTS_OUTPUT: str = "DeckLists"
 KEY_WORDS_OUTPUT: str = "Keywords"
+PRICES_OUTPUT: str = "Prices"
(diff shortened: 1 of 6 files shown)


ancestral:
I would say having one price for each card in the card files might be advantageous for any users with sites that have relied on the pricing. They might not care about the price history, but like having the current price.

That being said, MTGJSON hasn’t had pricing very long, so probably it wouldn’t be a big issue separating it. People who really want it can reference it in its own prices file.

PR #637, opened 2020-07-23
Add some missing French translations and fix 1 existing trnaslation

lionelh:


ZeldaZach (approved):
Thanks for your fix! It's greatly appreciated

PR #644, opened 2020-07-31
TCGPlayer Sku Data Added to Identifiers 

tpblaster:
I wrote a function to add TCGPlayer sku data to the identifiers class. The function uses the endpoint Get Product Details(https://docs.tcgplayer.com/v1.38.0/reference#catalog_getproduct-1) that passes product ids from existing data and retrieves sku ids. This function is run after product ids are retrieved and needs API keys from TCGPlayer to run, so it will need to be skipped if those are not configured.


mtgjson5/set_builder.py
```diff
@@ -1038,3 +1040,71 @@ def get_base_and_total_set_sizes(set_code: str) -> Tuple[int, int]:
...
+def add_tcgplayer_details(mtgjson_set: MtgjsonSetObject) -> MtgjsonSetObject:
+    """
+    Adds a list of all sku ids from TCGPlayer
+    :param mtgjson_set: MTGJSON Set
+    """
+
+    LOGGER.info(f"Adding TCGPlayer details for {mtgjson_set.code}")
+    list_of_product_ids: List
+    group_sku_data: Dict[str, List]
+    condition_map = {"1": "NM", "2": "LP", "3": "MP", "4": "HP", "5": "DMG", "6": "U"}
```
ZeldaZach: These two variables would make nice class variables in TCGPlayerProvider()

mtgjson5/set_builder.py
```diff
@@ -1038,3 +1040,71 @@ def get_base_and_total_set_sizes(set_code: str) -> Tuple[int, int]:
     total_set_size = len(ScryfallProvider().download_cards(set_code))
 
     return base_set_size, total_set_size
+
+
+def add_tcgplayer_details(mtgjson_set: MtgjsonSetObject) -> MtgjsonSetObject:
+    """
+    Adds a list of all sku ids from TCGPlayer
+    :param mtgjson_set: MTGJSON Set
+    """
```
ZeldaZach, line 1050: missing return type

mtgjson5/set_builder.py
```diff
@@ -1038,3 +1040,71 @@ def get_base_and_total_set_sizes(set_code: str) -> Tuple[int, int]:
...
+        "5": "GE",
+        "6": "IT",
+        "7": "JP",
+        "8": "KR",
+        "9": "PT",
+        "10": "RU",
+        "11": "SP",
+    }
+    for mtgjson_card in mtgjson_set.cards:
+        if mtgjson_card.identifiers.tcgplayer_product_id is None:
```
ZeldaZach: You can do `if not X:` for null comparisons

mtgjson5/set_builder.py
```diff
@@ -1038,3 +1040,71 @@ def get_base_and_total_set_sizes(set_code: str) -> Tuple[int, int]:
...
+        "4": "FR",
+        "5": "GE",
+        "6": "IT",
+        "7": "JP",
+        "8": "KR",
+        "9": "PT",
+        "10": "RU",
+        "11": "SP",
+    }
+    for mtgjson_card in mtgjson_set.cards:
```
ZeldaZach: I feel this entire for loop can be replaced with a list comprehension... maybe something like

```python
list_of_product_ids = [mtgjson_card.identifiers.tcgplayer_product_id for mtgjson_card in mtgjson_set.cards if mtgjson_card.identifiers.tcgplayer_product_id]
```

mtgjson5/set_builder.py
```diff
@@ -1038,3 +1040,71 @@ def get_base_and_total_set_sizes(set_code: str) -> Tuple[int, int]:
...
+        if mtgjson_card.identifiers.tcgplayer_product_id is None:
+            continue
+        list_of_product_ids.append(mtgjson_card.identifiers.tcgplayer_product_id)
+    for counter in range(0, len(list_of_product_ids), 250):
```
ZeldaZach: The 0 is implicit on range, so `range(len(list_of_product_ids), 250):` should be fine

mtgjson5/set_builder.py
```diff
@@ -1038,3 +1040,71 @@ def get_base_and_total_set_sizes(set_code: str) -> Tuple[int, int]:
...
+                condition = condition_map[str(sku["conditionId"])]
+                # Translated printing
+                if sku["printingId"] == 1:
+                    printing = "Normal"
+                else:
+                    printing = "Foil"
+                # Translates language
+                language = language_map[str(sku["languageId"])]
+                translated_skus.append(
+                    {
```
ZeldaZach: This is the kind of place I'd expect to see a different class. However, for identifiers I'd like to keep it as a string to string translation... I'm not sure the best approach here right now

mtgjson5/set_builder.py
```diff
@@ -1038,3 +1040,71 @@ def get_base_and_total_set_sizes(set_code: str) -> Tuple[int, int]:
...
+                product_id = str(sku["productId"])
+                # Translates condition
+                condition = condition_map[str(sku["conditionId"])]
+                # Translated printing
+                if sku["printingId"] == 1:
+                    printing = "Normal"
+                else:
+                    printing = "Foil"
+                # Translates language
+                language = language_map[str(sku["languageId"])]
```
ZeldaZach: I don't believe you have to stringify the languageId. If you do, maybe change the conversion table to int->str instead? (This comment can be applied to more variables here)

mtgjson5/set_builder.py
```diff
@@ -1038,3 +1040,71 @@ def get_base_and_total_set_sizes(set_code: str) -> Tuple[int, int]:
...
+                        "language": language,
+                        "printing": printing,
+                        "condition": condition,
+                    }
+                )
+            group_sku_data[product_id] = translated_skus
+            LOGGER.info(f"Adding TCGPlayer details for {translated_skus}")
+
+    for mtgjson_card in mtgjson_set.cards:
+        if mtgjson_card.identifiers.tcgplayer_product_id is None:
```
ZeldaZach: Instead, why not just check `if X: set sku_ids` or better yet, `sku_ids = group_sku_data.get(prod_id)` This will set it to None by default if the value isn't found

mtgjson5/set_builder.py
```diff
@@ -1038,3 +1040,71 @@ def get_base_and_total_set_sizes(set_code: str) -> Tuple[int, int]:
...
+            continue
+        mtgjson_card.identifiers.tcgplayer_sku_ids = group_sku_data[
+            mtgjson_card.identifiers.tcgplayer_product_id
+        ]
+
+    return mtgjson_set
```
ZeldaZach: For this function, you don't actually have to return anything, as we are modifying an object in-place. You can see the return value isn't used by your change above where you simply call `add_tcgplayer_details(set) instead of set = add_tcgplayer_details(set)` :)

mtgjson5/classes/mtgjson_identifiers.py
```diff
@@ -24,11 +24,13 @@ class MtgjsonIdentifiersObject:
     scryfall_oracle_id: Optional[str]
     tcgplayer_product_id: Optional[str]
     mtgjson_v4_id: Optional[str]
+    tcgplayer_sku_ids: List[Dict[str, str]]
```
ZeldaZach: Yeah we're gonna have to address this... @kodabb maybe you have some suggestions?

mtgjson5/classes/mtgjson_identifiers.py
```diff
@@ -24,11 +24,13 @@ class MtgjsonIdentifiersObject:
...
 
     def __init__(self) -> None:
         """
         Empty initializer
         """
+        self.tcgplayer_product_id = None
```
ZeldaZach: nitpick: I like setting vars alphabetically (so mid gets set before tcg)

mtgjson5/set_builder.py
ZeldaZach: (Should actually be None)

ZeldaZach (changes requested):
Thanks so much for your contribution! I left you some feedback that should be addressed before I get around to testing. If you have any Qs, don't hesitate to reach out.

mtgjson5/classes/mtgjson_identifiers.py
kodawah: I'd honestly move these details to a separate dictionary, so that you don't pollute `identifiers` with too much information and can have any format for the new entry

I would propose `tcgplayer_sku[string]{tcgid, language, condition}` at the card root level (aka same as identifiers)

example

    "tcgplayer_skus" : {
        "3455" : {
            "product_id": "1348"
            "language": "EN",
            "condition": "LP",
            ...
        }
    }

mtgjson5/providers/tcgplayer.py
```diff
@@ -24,6 +24,20 @@ class TCGPlayerProvider(AbstractProvider):
     """
 
     api_version: str = ""
+    condition_map = {1: "NM", 2: "LP", 3: "MP", 4: "HP", 5: "DMG", 6: "U"}
```
ZeldaZach: I'd like to see these values built at run time instead of static, just in case they change something on their API. 
See https://docs.tcgplayer.com/reference#catalog_getconditions should get condition map. Language map, I'm not sure at this exact moment.

ChrisGallardo (approved):

ZeldaZach:
This is closing as a no-addition for the time being, as I don't have a great way of sharing this data without breaking a lot of structs. Will re-evaulate in the future.

PR #646, opened 2020-08-10
TCGPlayer Buylist Data Standalone

tpblaster:
This feature uses the TCGPlayer API to generate MTGJSON price objects from TCGPlayer buylist info. Using the List All Products API endpoint the system builds a map of each product id to the near mint foil and nonfoil sku. Then using List Product Buylist Prices by Group endpoint all potential sku buy prices are collected. Then the near mint foil and nonfoil skus are saved to the MTGJSON price objects. This could be made more efficient in the future by storing the sku data locally in the allprintings file.

### DIFF
--- a/mtgjson5/price_builder.py
+++ b/mtgjson5/price_builder.py
@@ -21,6 +21,7 @@ from .providers import (
     CardMarketProvider,
     TCGPlayerProvider,
 )
+from .utils import deep_merge_dictionaries
 
 LOGGER = logging.getLogger(__name__)
 
@@ -133,30 +134,6 @@ def prune_prices_archive(content: Dict[str, Any], months: int = 3) -> None:
     LOGGER.info(f"Pruned {keys_pruned} structs")
 
 
-def deep_merge_dictionaries(
-    dictionary_one: Dict[str, Any], dictionary_two: Dict[str, Any]
-) -> Dict[str, Any]:
-    """
-    Merge two dictionaries together, recursively
-    :param dictionary_one: Dict 1
-    :param dictionary_two: Dict 2
-    :return: Combined Dictionaries
-    """
-    new_dictionary = dictionary_one.copy()
-
-    new_dictionary.update(
-        {
-            key: deep_merge_dictionaries(new_dictionary[key], dictionary_two[key])
-            if isinstance(new_dictionary.get(key), dict)
-            and isinstance(dictionary_two[key], dict)
-            else dictionary_two[key]
-            for key in dictionary_two.keys()
-        }
-    )
-
-    return new_dictionary
-
-
 def build_today_prices() -> Dict[str, Any]:
     """
     Get today's prices from upstream sources and combine them together
@@ -170,14 +147,10 @@ def build_today_prices() -> Dict[str, Any]:
     tcgplayer = _generate_prices(TCGPlayerProvider())
     card_market = _generate_prices(CardMarketProvider())
     card_kingdom = _generate_prices(CardKingdomProvider())
-    tcgplayer_buylist = _generate_tcgplayer_buylist_prices(TCGPlayerProvider())
 
     final_results = deep_merge_dictionaries(
-        deep_merge_dictionaries(
-            deep_merge_dictionaries(card_hoarder, tcgplayer),
-            deep_merge_dictionaries(card_market, card_kingdom),
-        ),
-        tcgplayer_buylist,
+        deep_merge_dictionaries(card_hoarder, tcgplayer),
+        deep_merge_dictionaries(card_market, card_kingdom),
     )
 
     return final_results
@@ -191,21 +164,6 @@ def _generate_prices(provider: Any) -> Dict[str, Any]:
     :return Manageable data for MTGJSON prices
     """
     preprocess_prices = provider.generate_today_price_dict(
-        OUTPUT_PATH.joinpath("AllPrintings.json")
-    )
-    final_prices: Dict[str, Any] = json.loads(
-        json.dumps(preprocess_prices, default=lambda o: o.to_json())
-    )
-    return final_prices
-
-
-def _generate_tcgplayer_buylist_prices(provider: TCGPlayerProvider) -> Dict[str, Any]:
-    """
-    Generates TCGPlayer buylist prices and preps them for merging
-    :param provider: MTGJSON Provider that implements generate_today_buylist_price_dict
-    :return Manageable data for MTGJSON prices
-    """
-    preprocess_prices = provider.generate_today_buylist_price_dict(
         OUTPUT_PATH.joinpath("AllPrintings.json")
     )
     final_prices: Dict[str, Any] = json.loads(
@@ -277,9 +235,7 @@ def build_prices() -> Dict[str, Any]:
     today_prices = build_today_prices()
 
     if not today_prices:
-        LOGGER.warning(
-            "TCGPlayer and CardHoarder keys not established. No prices generated"
-        )
+        LOGGER.warning("Pricing information failed to generate")
         return {}
 
     archive_prices = get_price_archive_data()
--- a/mtgjson5/price_builder.py
+++ b/mtgjson5/price_builder.py
@@ -170,10 +170,14 @@ def build_today_prices() -> Dict[str, Any]:
     tcgplayer = _generate_prices(TCGPlayerProvider())
     card_market = _generate_prices(CardMarketProvider())
     card_kingdom = _generate_prices(CardKingdomProvider())
+    tcgplayer_buylist = _generate_tcgplayer_buylist_prices(TCGPlayerProvider())
 
     final_results = deep_merge_dictionaries(
-        deep_merge_dictionaries(card_hoarder, tcgplayer),
-        deep_merge_dictionaries(card_market, card_kingdom),
+        deep_merge_dictionaries(
+            deep_merge_dictionaries(card_hoarder, tcgplayer),
+            deep_merge_dictionaries(card_market, card_kingdom),
+        ),
+        tcgplayer_buylist,
     )
 
     return final_results
@@ -187,6 +191,21 @@ def _generate_prices(provider: Any) -> Dict[str, Any]:
     :return Manageable data for MTGJSON prices
     """
     preprocess_prices = provider.generate_today_price_dict(
+        OUTPUT_PATH.joinpath("AllPrintings.json")
+    )
+    final_prices: Dict[str, Any] = json.loads(
+        json.dumps(preprocess_prices, default=lambda o: o.to_json())
+    )
+    return final_prices
+
+
+def _generate_tcgplayer_buylist_prices(provider: TCGPlayerProvider) -> Dict[str, Any]:
+    """
+    Generates TCGPlayer buylist prices and preps them for merging
+    :param provider: MTGJSON Provider that implements generate_today_buylist_price_dict
+    :return Manageable data for MTGJSON prices
+    """
+    preprocess_prices = provider.generate_today_buylist_price_dict(
         OUTPUT_PATH.joinpath("AllPrintings.json")
     )
     final_prices: Dict[str, Any] = json.loads(
--- a/mtgjson5/providers/tcgplayer.py
+++ b/mtgjson5/providers/tcgplayer.py
@@ -10,7 +10,7 @@ import requests
 from singleton_decorator import singleton
 
 from ..classes import MtgjsonPricesObject
-from ..consts import CACHE_PATH, OUTPUT_PATH
+from ..consts import CACHE_PATH
 from ..providers.abstract import AbstractProvider
 from ..utils import parallel_call, retryable_session
 
@@ -56,8 +56,8 @@ class TCGPlayerProvider(AbstractProvider):
             return ""
 
         if not (
-            config.get("TCGPlayer", "client_id")
-            and config.get("TCGPlayer", "client_secret")
+                config.get("TCGPlayer", "client_id")
+                and config.get("TCGPlayer", "client_secret")
         ):
             LOGGER.warning("TCGPlayer keys not established. Skipping requests")
             self.__keys_found = False
@@ -128,7 +128,7 @@ class TCGPlayerProvider(AbstractProvider):
         return magic_set_ids
 
     def generate_today_price_dict(
-        self, all_printings_path: pathlib.Path
+            self, all_printings_path: pathlib.Path
     ) -> Dict[str, MtgjsonPricesObject]:
         """
         Download the TCGPlayer pricing API and collate into MTGJSON format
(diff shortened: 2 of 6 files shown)


mtgjson5/providers/tcgplayer.py
```diff
@@ -148,6 +148,71 @@ def generate_today_price_dict(
...
+        if not api_response:
+            # No more entries
+            break
+
+        response = json.loads(api_response)
+        if not response["results"]:
+            # Something went wrong
+            break
+
+        for product_data in response["results"]:
```
ZeldaZach: can simply say magic_set_product_data.extend(response["results"])

mtgjson5/providers/tcgplayer.py
```diff
@@ -148,6 +148,71 @@ def generate_today_price_dict(
...
+            magic_set_product_data.append(product_data)
+
+        api_offset += len(response["results"])
+
+    return magic_set_product_data
+
+
+def generate_tcgplayer_sku_map(tcgplayer_set_sku_data: List[Dict]) -> Dict[str, List]:
```
ZeldaZach: Your types don't make sense here. For non-primatives, you'll need to be more detailed. List[Dict[str, str]] for example. You'll need to figure out your types.

mtgjson5/providers/tcgplayer.py
```diff
@@ -148,6 +148,71 @@ def generate_today_price_dict(
...
+    takes product info and builds a sku map
+    :param tcgplayer_set_sku_data: list of product data dicts used to a build a product id to sku map
+    :return: Map of TCGPlayerID -> NM Foil and Nonfoil SKU
+    """
+    tcgplayer_sku_map: Dict[str, List] = {}
+    for product_data in tcgplayer_set_sku_data:
+        nonfoil_sku: int = 0
+        foil_sku: int = 0
+        for sku in product_data["skus"]:
+            if (
```
ZeldaZach: Might be overkill, but ENUM classes would be very helpful here to better identify what these values and numbers mean

mtgjson5/providers/tcgplayer.py
```diff
@@ -148,6 +148,71 @@ def generate_today_price_dict(
...
+                and sku["languageId"] == 1
+            ):
+                nonfoil_sku = sku["skuId"]
+            elif (
+                sku["conditionId"] == 1
+                and sku["printingId"] == 2
+                and sku["languageId"] == 1
+            ):
+                foil_sku = sku["skuId"]
+        tcgplayer_sku_map[str(product_data["productId"])] = [nonfoil_sku, foil_sku]
```
ZeldaZach: If you're returning a specific ordering, I'd consider making it a dict instead. {"foil": ..., "nonfoil": ...} for example.

mtgjson5/providers/tcgplayer.py
```diff
@@ -171,6 +236,62 @@ def generate_tcgplayer_to_mtgjson_map(
...
+    """
+    LOGGER.info(f"Building {group_id_and_name[1]} TCGPlayer Buylist data")
+    api_response = TCGPlayerProvider().download(
+        f"https://api.tcgplayer.com/pricing/buy/group/{group_id_and_name[0]}"
+    )
+    # breaks if the request is bad
+    if not api_response:
+        return {}
+    response = json.loads(api_response)
+    # breaks if there is no data
```
ZeldaZach: unneeded

mtgjson5/providers/tcgplayer.py
```diff
@@ -171,6 +236,62 @@ def generate_tcgplayer_to_mtgjson_map(
...
+    """
+    takes a group id and name and finds all buylist data for that group
+    :param group_id_and_name: TCGPlayer Set ID & Name to build
+    :return: returns a map of tcgplayer buylist data to card uuids
+    """
+    LOGGER.info(f"Building {group_id_and_name[1]} TCGPlayer Buylist data")
+    api_response = TCGPlayerProvider().download(
+        f"https://api.tcgplayer.com/pricing/buy/group/{group_id_and_name[0]}"
+    )
+    # breaks if the request is bad
```
ZeldaZach: unneeded

mtgjson5/providers/tcgplayer.py
```diff
@@ -171,6 +236,62 @@ def generate_tcgplayer_to_mtgjson_map(
...
+        OUTPUT_PATH.joinpath("AllPrintings.json")
+    )
+    # this is set specific, built each time unlike the uuid map
+    sku_map = generate_tcgplayer_sku_map(get_tcgplayer_sku_data(group_id_and_name))
+    for product_buylist_data in response["results"]:
+        # checks if the product id is in the uuid map
+        if not uuid_map.get(str(product_buylist_data["productId"])):
+            continue
+        # if it is in the uuid map we set the key for the pricing object
+        key: str = str(uuid_map.get(str(product_buylist_data["productId"])))
```
ZeldaZach: You don't need to define types for objects always. I'd also look at the uuid_map to make sure it's a str->str map as there's no reason to hard convert

mtgjson5/providers/tcgplayer.py
```diff
@@ -171,6 +236,62 @@ def generate_tcgplayer_to_mtgjson_map(
...
+                            "paper", "tcgplayer", TCGPlayerProvider().today_date
+                        )
+
+                    prices_map[key].buy_normal = card_price
+            # checks if the sku is NM Foil
+            elif sku["skuId"] == sku_map[str(product_buylist_data["productId"])][1]:
+                if sku["prices"]["high"] is not None:
+                    card_price = sku["prices"]["high"]
+                    if key not in prices_map.keys():
+                        prices_map[key] = MtgjsonPricesObject(
```
ZeldaZach: Take a closer look at the if/elseif you wrote. If A: (stuff) normal = price elif B: (stuff) foil = price. With this, you can pull (stuff) outside the loops and reduce code complexity. If you're copy/pasting more than 1 line in code, you can (probably) refactor to avoid it :)

mtgjson5/providers/tcgplayer.py
```diff
@@ -214,3 +335,26 @@ def get_tcgplayer_prices_map(
...
+    across the system.
+    :param all_printings_path: Path to JSON compiled version
+    :return: Map of TCGPlayerID -> MTGJSON UUID
+    """
+    with all_printings_path.expanduser().open(encoding="utf-8") as f:
+        file_contents = json.load(f).get("data", {})
+
+    dump_map: Dict[str, str] = {}
+    for value in file_contents.values():
+        for card in value.get("cards", []) + value.get("tokens", []):
```
ZeldaZach, line 419: I swear I wrote a function to redux this change. Might have been on your other PR.

mtgjson5/providers/tcgplayer.py
```diff
@@ -148,6 +148,71 @@ def generate_today_price_dict(
...
+
+def generate_tcgplayer_sku_map(tcgplayer_set_sku_data: List[Dict]) -> Dict[str, List]:
+    """
+    takes product info and builds a sku map
+    :param tcgplayer_set_sku_data: list of product data dicts used to a build a product id to sku map
+    :return: Map of TCGPlayerID -> NM Foil and Nonfoil SKU
+    """
+    tcgplayer_sku_map: Dict[str, List] = {}
+    for product_data in tcgplayer_set_sku_data:
+        nonfoil_sku: int = 0
```
ZeldaZach: I'd consider Nullifying these instead of 0's (Optional[int]) as it's more clear imo

mtgjson5/providers/tcgplayer.py
```diff
@@ -171,6 +236,62 @@ def generate_tcgplayer_to_mtgjson_map(
...
+        # checks if the product id is in the uuid map
+        if not uuid_map.get(str(product_buylist_data["productId"])):
+            continue
+        # if it is in the uuid map we set the key for the pricing object
+        key: str = str(uuid_map.get(str(product_buylist_data["productId"])))
+        # parse each sku to find the near mint skus
+        for sku in product_buylist_data["skus"]:
+            # checks if the sku is NM Nonfoil
+            if sku["skuId"] == sku_map[str(product_buylist_data["productId"])][0]:
+                if sku["prices"]["high"] is not None:
```
ZeldaZach: This can be simplified to "if A": instead of "if A is not None" (unless we're specifically only looking for a None and will accept Noneable stuff

mtgjson5/providers/tcgplayer.py
```diff
@@ -148,6 +148,75 @@ def generate_today_price_dict(
         return dict(results)
 
 
+def get_tcgplayer_sku_data(group_id_and_name: Tuple[str, str]) -> List[Dict]:
```
ZeldaZach: Return type is still funky

mtgjson5/providers/tcgplayer.py
```diff
@@ -148,6 +148,75 @@ def generate_today_price_dict(
...
+    """
+    finds all sku data for a given group using the TCGPlayer API
```
ZeldaZach: Nit: F should be cap

mtgjson5/providers/tcgplayer.py
```diff
@@ -148,6 +148,75 @@ def generate_today_price_dict(
...
+            ):
+                nonfoil_sku = sku["skuId"]
+            elif (
+                sku["conditionId"] == 1
+                and sku["printingId"] == 2
+                and sku["languageId"] == 1
+            ):
+                foil_sku = sku["skuId"]
+        tcgplayer_sku_map[str(product_data["productId"])] = {
+            "nonfoilSku": nonfoil_sku,
```
ZeldaZach: I'd prefer to see these as snake_case than camelCase because Python

mtgjson5/providers/tcgplayer.py
```diff
@@ -171,6 +240,61 @@ def generate_tcgplayer_to_mtgjson_map(
...
+    response = json.loads(api_response)
+    if not response["results"]:
+        return {}
+    prices_map: Dict[str, MtgjsonPricesObject] = {}
+    # when building the full system, only build this once and take it as an arg/write to a file
+    uuid_map = generate_tcgplayer_to_mtgjson_map(
+        OUTPUT_PATH.joinpath("AllPrintings.json")
+    )
+    # this is set specific, built each time unlike the uuid map
+    sku_map = generate_tcgplayer_sku_map(get_tcgplayer_sku_data(group_id_and_name))
```
ZeldaZach: Function to a function like this is a bit yucky. I'd try to clarify what's going on here

mtgjson5/providers/tcgplayer.py
```diff
@@ -171,6 +240,61 @@ def generate_tcgplayer_to_mtgjson_map(
     return dump_map
 
 
+def get_tcgplayer_buylist_prices_map(
```
ZeldaZach, line 307: Nothing seems to call this function?

mtgjson5/providers/tcgplayer.py
```diff
@@ -171,6 +240,61 @@ def generate_tcgplayer_to_mtgjson_map(
...
+    group_id_and_name: Tuple[str, str]
```
ZeldaZach: Also, why a tuple? This is prime for `group_id: str, group_name: str` which is cleaner

ZeldaZach: ^ That goes for everywhere you used a tuple btw

mtgjson5/providers/tcgplayer.py
ZeldaZach: This is still a problem as you try to make the price object twice. I'd do this as 

for product in response:
   if product id not found continue
   if key not found: make prices object
   for sku in skus:
      if prices.high: if (nonfoil) setNonFoil else setFoil

mtgjson5/providers/tcgplayer.py
```diff
@@ -147,6 +147,123 @@ def generate_today_price_dict(
...
+        """
+        Download the TCGPlayer pricing API and collate into MTGJSON format
+        :param all_printings_path Path to AllPrintings.json for pre-processing
+        :return: Buylist Prices to combine with others
+        """
+
+        if not self.__keys_found:
+            LOGGER.warning("Keys not found for TCGPlayer, skipping")
+            return {}
+        uuid_map = generate_tcgplayer_to_mtgjson_map(all_printings_path)
```
ZeldaZach: missing line break

mtgjson5/providers/tcgplayer.py
```diff
@@ -147,6 +147,123 @@ def generate_today_price_dict(
...
+        self, all_printings_path: pathlib.Path
+    ) -> List[Dict[str, MtgjsonPricesObject]]:
+        """
+        Download the TCGPlayer pricing API and collate into MTGJSON format using a sequential process
+        :param all_printings_path Path to AllPrintings.json for pre-processing
+        :return: Buylist Prices to combine, not in correct format yet
+        """
+        if not self.__keys_found:
+            LOGGER.warning("Keys not found for TCGPlayer, skipping")
+            return [{}]
```
ZeldaZach: Can simply return []

mtgjson5/providers/tcgplayer.py
```diff
@@ -171,6 +288,68 @@ def generate_tcgplayer_to_mtgjson_map(
...
+        key: str = uuid_map[str(product_buylist_data["productId"])]
+        # parse each sku to find the near mint skus
+        for sku in product_buylist_data["skus"]:
+            if sku["prices"]["high"]:
+                if sku_map.get(str(product_buylist_data["productId"])):
+                    if sku["skuId"] == sku_map[
+                        str(product_buylist_data["productId"])
+                    ].get("nonfoilSku"):
+
+                        if key not in prices_map.keys():
```
ZeldaZach: Still not happy with the dup code

mtgjson5/providers/tcgplayer.py
```diff
@@ -147,6 +147,123 @@ def generate_today_price_dict(
...
+            LOGGER.warning("Keys not found for TCGPlayer, skipping")
+            return [{}]
+        uuid_map = generate_tcgplayer_to_mtgjson_map(all_printings_path)
+        CACHE_PATH.mkdir(parents=True, exist_ok=True)
+        with CACHE_PATH.joinpath("tcgplayer_buylist_price_map.json").open("w") as file:
+            json.dump(uuid_map, file)
+
+        ids_and_names = self.get_tcgplayer_magic_set_ids()
+        set_price_data = []
+        for set_ids in ids_and_names:
```
ZeldaZach: Could use list comp here.
set_price_data = [get_tcg(set_ids) for set_ids in ids_and_names]

ZeldaZach: Also your variable names don't match your function name. "ids_and_names" when the function says get_set_ids only...?

mtgjson5/price_builder.py
```diff
@@ -159,10 +159,14 @@ def build_today_prices() -> Dict[str, Any]:
     tcgplayer = _generate_prices(TCGPlayerProvider())
     card_market = _generate_prices(CardMarketProvider())
     card_kingdom = _generate_prices(CardKingdomProvider())
+    tcgplayer_buylist = _generate_tcgplayer_buylist_prices(TCGPlayerProvider())
```
ZeldaZach: I don't agree with making a new function. We have an internal API (read: interface) for a reason. If you could make it fit that model, would be more likable

mtgjson5/price_builder.py
```diff
@@ -159,10 +159,14 @@ def build_today_prices() -> Dict[str, Any]:
...
 
     final_results = deep_merge_dictionaries(
-        deep_merge_dictionaries(card_hoarder, tcgplayer),
-        deep_merge_dictionaries(card_market, card_kingdom),
+        deep_merge_dictionaries(
```
ZeldaZach: Would like to see DMD rewritten to take any amount of dicts... might ticket this for myself

mtgjson5/price_builder.py
```diff
@@ -184,6 +188,21 @@ def _generate_prices(provider: Any) -> Dict[str, Any]:
     return final_prices
 
 
+def _generate_tcgplayer_buylist_prices(provider: TCGPlayerProvider) -> Dict[str, Any]:
```
ZeldaZach: Should not exist

mtgjson5/providers/tcgplayer.py
```diff
@@ -147,6 +147,124 @@ def generate_today_price_dict(
...
+
+        ids_and_names = self.get_tcgplayer_magic_set_ids()
+
+        results = parallel_call(
+            get_tcgplayer_buylist_prices_map, ids_and_names, fold_dict=True
+        )
+
+        return dict(results)
+
+    def generate_buylist_price_sequential(
```
ZeldaZach: Unused?

mtgjson5/providers/tcgplayer.py
```diff
@@ -147,6 +147,124 @@ def generate_today_price_dict(
...
+    ) -> List[Dict[str, MtgjsonPricesObject]]:
+        """
+        Download the TCGPlayer pricing API and collate into MTGJSON format using a sequential process
+        :param all_printings_path Path to AllPrintings.json for pre-processing
+        :return: Buylist Prices to combine, not in correct format yet
+        """
+        if not self.__keys_found:
+            LOGGER.warning("Keys not found for TCGPlayer, skipping")
+            return []
+        uuid_map = generate_tcgplayer_to_mtgjson_map(all_printings_path)
```
ZeldaZach: line break

mtgjson5/providers/tcgplayer.py
```diff
@@ -147,6 +147,124 @@ def generate_today_price_dict(
...
+            # No more entries
+            break
+
+        response = json.loads(api_response)
+        if not response["results"]:
+            # Something went wrong
+            break
+
+        magic_set_product_data.extend(response["results"])
+
```
ZeldaZach: remove extra space line plz

mtgjson5/providers/tcgplayer.py
```diff
@@ -147,6 +147,124 @@ def generate_today_price_dict(
...
+                and sku["languageId"] == 1
+            ):
+                nonfoil_sku = sku["skuId"]
+            elif (
+                sku["conditionId"] == 1
+                and sku["printingId"] == 2
+                and sku["languageId"] == 1
+            ):
+                foil_sku = sku["skuId"]
+        tcgplayer_sku_map[str(product_data["productId"])] = {
```
ZeldaZach: I'd honestly be inclined to not include NULL values into this dict, and use the .get() method downstream... would remove some complexity in types

ZeldaZach:
Tested and works as expected. Memory graph as follows:
memory-profile

ZeldaZach (approved):
Thanks for setting the grassroots for me. Glad we could get this into MTGJSON :)

ZeldaZach:
cc @omfgitsmark  @kodabb for a glance if you're interested

kodawah (approved):

ZeldaZach:
Thanks for setting us up for additional success. Buylist prices will appear starting tomorrow :)

Issue #530, opened 2020-03-13
Will Kenrith and Rowan Kenrith foil-only


kodawah:
Would it be possible to add some kind of flag to distinguish these two planeswalkers? In Battlebond they are numbered 1/255 and 2/256, and there is no way to distinguish them apart.
I believe just adding a IsAlternative=true would be enough, barring some additional fields in Set as suggested here: https://github.com/mtgjson/mtgjson/issues/525

kodawah:
confirmed, adding isAlternative is all that is needed, it's a similar case like kaya, ghost assassin

PR #653, opened 2020-08-14
Tag JMP and 2XM extra cards with IsAlternative

### LINKED ISSUE #652: set isAlternate for cards with same name and same frame in JMP and 2XM
#### Version(s) affected
- [x] MTGJSON 4
- [x] MTGJSON 5

#### Description of Bug
similar to #530 we need to set some IsAlternate properties for some cards that are duplicated in the new sets, namely
* Scholar of the Lost Trove JMP 496
* Wrath of God 2XM 383
* Chord of Calling 2XM 384

kodawah:
Fix #652.

Untested, but it's the same at #530. I think it's fine for JMP, but I don't know is setting the 2XM limit so low has the negative effect of setting box topper cards as alternatives which i think it's not correct.


ZeldaZach:
Hey @kodabb thanks for the contribution! I went with a simpler approach (c05fe3335b6a330cafd0d4f2ad422f0e63846870) to get this solved.

PR #658, opened 2020-08-19
Add AllIdentifiers.json

### LINKED ISSUE #650: add uuid->cardobject+ json file
It would be nice to have a json file to tie in allprintings and atomiccards, so that it is possible to map a uuid to a cardobject. 

The main caveat is that each entry would need a SetCode and/or edition name to retrieve the set properties

ZeldaZach:
Fix #650 by ADDING AllIdentifiers.json which is UUID -> Card Object, where Card Object also gets a setCode attribute for better backwards compatibility.

cc: @kodabb

### DIFF
--- a/mtgjson5/compiled_classes/__init__.py
+++ b/mtgjson5/compiled_classes/__init__.py
@@ -2,6 +2,7 @@
 MTGJSON Compiled Outputs Dispatcher
 """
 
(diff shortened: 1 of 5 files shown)


kodawah:
looks ok to me, thanks for taking the time to doing it @ZeldaZach

ZeldaZach:
cc for docs @staghouse

PR #670, opened 2020-09-03
Fix #668 added AllTcgSkus to compiled outputs 

### LINKED ISSUE #668: compile a new file for tcg skus
So i was thinking of a way to integrate @tpblaster's changes from https://github.com/mtgjson/mtgjson/pull/644

If we want to avoid including vendor-specific information in the main files, maybe we could shove vendor specific data in a separate file. I'd design AllTCGSkus.json like AllIdentifiers, so a map of uuids pointing to the map of sku and product id

The only question is if we need to something similar for mkm too

tpblaster:
I wrote a new class to build a compiled output that maps a UUID to the TCGPlayer sku data for that card. This class may need some tweaking to better mirror how AllIdentifiers handles the usage of AllPrintings. If this pull request gets approved, the way TCGPlayer buylist data is compiled could be reworked to be more efficient.

### DIFF
--- a/mtgjson5/compiled_classes/__init__.py
+++ b/mtgjson5/compiled_classes/__init__.py
@@ -4,7 +4,7 @@ MTGJSON Compiled Outputs Dispatcher
 
 from .mtgjson_all_identifiers import MtgjsonAllIdentifiersObject
 from .mtgjson_all_printings import MtgjsonAllPrintingsObject
-from .mtgjson_all_tcg_skus import MtgjsonAllTcgSkusObject
+from .mtgjson_all_tcg_skus import MtgjsonAllTcgplayerSkusObject
 from .mtgjson_atomic_cards import MtgjsonAtomicCardsObject
 from .mtgjson_card_types import MtgjsonCardTypesObject
 from .mtgjson_compiled_list import MtgjsonCompiledListObject
--- a/mtgjson5/compiled_classes/__init__.py
+++ b/mtgjson5/compiled_classes/__init__.py
@@ -4,6 +4,7 @@ MTGJSON Compiled Outputs Dispatcher
 
 from .mtgjson_all_identifiers import MtgjsonAllIdentifiersObject
 from .mtgjson_all_printings import MtgjsonAllPrintingsObject
+from .mtgjson_all_tcg_skus import MtgjsonAllTcgSkusObject
 from .mtgjson_atomic_cards import MtgjsonAtomicCardsObject
 from .mtgjson_card_types import MtgjsonCardTypesObject
 from .mtgjson_compiled_list import MtgjsonCompiledListObject
--- a/mtgjson5/compiled_classes/__init__.py
+++ b/mtgjson5/compiled_classes/__init__.py
@@ -4,7 +4,6 @@ MTGJSON Compiled Outputs Dispatcher
 
 from .mtgjson_all_identifiers import MtgjsonAllIdentifiersObject
 from .mtgjson_all_printings import MtgjsonAllPrintingsObject
-from .mtgjson_all_tcg_skus import MtgjsonAllTcgplayerSkusObject
 from .mtgjson_atomic_cards import MtgjsonAtomicCardsObject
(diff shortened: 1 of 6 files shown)


mtgjson5/compiled_classes/mtgjson_all_tcg_skus.py
```diff
@@ -0,0 +1,47 @@
+"""
+MTGJSON AllTcgSkus Object
+"""
+
+import logging
+import pathlib
+from typing import Dict
+
+from mtgjson5.providers import TCGPlayerProvider
```
ZeldaZach: These should be indirect import statements

mtgjson5/compiled_classes/mtgjson_structures.py
```diff
@@ -66,6 +68,7 @@ def __init__(self) -> None:
         self.referral_database = "ReferralMap"
         self.version = "Meta"
         self.all_identifiers = "AllIdentifiers"
+        self.all_tcg_skus = "AllTcgSkus"
```
ZeldaZach: I'd probably call this file TcgplayerSkus

mtgjson5/providers/tcgplayer.py
```diff
@@ -317,7 +317,7 @@ def get_tcgplayer_buylist_prices_map(
 
         for sku in buylist_data["skus"]:
             if not sku["prices"]["high"]:
-                # We only want to deal with the buyer paying the most for NM
+                # We want the buylist high price, not buylist market price
```
ZeldaZach, line 320: Unrelated, but i'll allow it

mtgjson5/compiled_classes/mtgjson_all_tcg_skus.py
ZeldaZach: (from ..providers, for example. Instead of from mtgjson5.providers)

mtgjson5/output_generator.py
```diff
@@ -19,6 +19,7 @@
     MtgjsonSetListObject,
     MtgjsonStructuresObject,
 )
+from .compiled_classes.mtgjson_all_tcg_skus import MtgjsonAllTcgSkusObject
```
ZeldaZach: This should be imported directly from compiled_classes, did you add this import to __init__ in the compiled_classes dir?

mtgjson5/output_generator.py
```diff
@@ -206,6 +207,13 @@ def build_all_printings_files(pretty_print: bool) -> None:
         pretty_print,
     )
 
+    # AllTcgSkus.json
```
ZeldaZach: Was this profiled? I split up these calls intentionally to avoid breaking the 3.5GiB Mem usage. This might be better suited elsewhere.

mtgjson5/compiled_classes/__init__.py
```diff
@@ -4,6 +4,7 @@
 
 from .mtgjson_all_identifiers import MtgjsonAllIdentifiersObject
 from .mtgjson_all_printings import MtgjsonAllPrintingsObject
+from .mtgjson_all_tcg_skus import MtgjsonAllTcgSkusObject
```
ZeldaZach: would rename to MtgjsonAllTcgplayerSkusObject (and renamed down the line accordingly)

mtgjson5/providers/tcgplayer.py
```diff
@@ -65,6 +65,28 @@ class TCGPlayerProvider(AbstractProvider):
 
     api_version: str = ""
     tcg_to_mtgjson_map: Dict[str, str]
+    language_map: Dict[int, str] = {
+        1: "English",
```
ZeldaZach: Since these are dicts with incremental keys, you could easily replace these lookups.

Now: word = language_map[key_id]
Next: word = language_options[key_id - 1]

ZeldaZach: Change of plans -- i'm gonna push a fix. Realized this is reinventing the wheel

PR #678, opened 2020-09-11
Fix #643 Added list of all printed languages to MTGJSON Set

### LINKED ISSUE #643: Available languages for printed sets
Hello! A nice little data set to have would be to add the available printed languages for a set. What I mean by this is for example, Masters 25 (A25) was only released in English, Japanese, and Simplified Chinese. There are only translations for set names, but nothing currently about this topic.

tpblaster:
This pull request adds a list of all printed non-English languages to the MTGJSON Set object. This list will be empty if the only available language for a set is English. This data is pulled from a sets card objects which each contain a MtgjsonForeignDataObject.


mtgjson5/set_builder.py
```diff
@@ -418,6 +423,19 @@ def build_mtgjson_set(set_code: str) -> Optional[MtgjsonSetObject]:
...
 
 
+def build_foreign_printed_languages(card_data: List[MtgjsonCardObject]) -> List[str]:
+    """
+    Constructs list of non-english languages that a set is printed in
+    :param card_data: all cards from that set
+    :return: a list of all non-english languages that a set is printed in
+    """
+    language_set: Set[str] = set()
+    for card in card_data:
```
ZeldaZach: The variable names in this function are a bit iffy imo, it's hard to follow.

I'd recommend taking "cards" as your arg, and "entry" instead of lang

mtgjson5/set_builder.py
```diff
@@ -401,6 +401,11 @@ def build_mtgjson_set(set_code: str) -> Optional[MtgjsonSetObject]:
     add_mcm_details(mtgjson_set)
     add_card_kingdom_details(mtgjson_set)
 
+    # Build Foreign Printed Languages list
```
ZeldaZach: Comment doesn't add much value

mtgjson5/classes/mtgjson_set.py
```diff
@@ -19,6 +19,7 @@ class MtgjsonSetObject:
     cards: List[MtgjsonCardObject]
     code: str
     code_v3: str
+    foreign_printed_languages: List[str]
```
ZeldaZach: I'm really thinking we want to support English too... As some sets in the future will be non-english, and then it'll get confusing. Idk, just thinking right now. Might want to just change this to printed_languages.

mtgjson5/providers/scryfall.py
```diff
@@ -29,6 +29,25 @@ class ScryfallProvider(AbstractProvider):
     CARDS_IN_SET: str = "https://api.scryfall.com/cards/search?order=set&q=set:{0}%20unique:prints"
     TYPE_CATALOG: str = "https://api.scryfall.com/catalog/{0}"
     cards_without_limits: Set[str]
+    language_map: Dict[str, str] = {
```
ZeldaZach: This is defined in consts.py as `LANGUAGE_MAP` already. No need to re implement.

mtgjson5/set_builder.py
```diff
@@ -650,7 +666,9 @@ def build_mtgjson_card(
     mtgjson_card.hand = scryfall_object.get("hand_modifier")
     mtgjson_card.has_foil = scryfall_object.get("foil")
     mtgjson_card.has_non_foil = scryfall_object.get("nonfoil")
-
+    mtgjson_card.language = ScryfallProvider().language_map.get(
```
ZeldaZach: Should use consts map (as i stated above)

mtgjson5/classes/mtgjson_set.py
```diff
@@ -19,6 +19,7 @@ class MtgjsonSetObject:
     cards: List[MtgjsonCardObject]
     code: str
     code_v3: str
+    printed_languages: List[str]
```
ZeldaZach, line 22: I don't know how true this is, as some single cards might have a foreign print while the other cards don't

mtgjson5/set_builder.py
```diff
@@ -418,6 +420,20 @@ def build_mtgjson_set(set_code: str) -> Optional[MtgjsonSetObject]:
...
+    Constructs list of non-english languages that a set is printed in
+    :param cards: all cards from that set
+    :return: a list of all non-english languages that a set is printed in
+    """
+    language_set: Set[str] = set()
+    for card in cards:
+        language_set.add(card.language)
+        for entry in card.foreign_data:
+            language_set.add(entry.language)
+    return list(language_set)
```
ZeldaZach, line 434: should also be sorted

ZeldaZach:
Closing for now, we can re-eval at some point with an up to date fork

PR #685, opened 2020-09-19
Add originalReleaseDate to CardObject

### LINKED ISSUE #683: add .originalReleaseDate to CardObject
Sometimes cards are added at a different point in time than the set.ReleaseDate, however this information is lost in the pulling data from upstream. It's important to keep this information if we want to sort cards and sets by release date

kodawah:
only with limited testing, fix #683

### DIFF
--- a/mtgjson5/set_builder.py
+++ b/mtgjson5/set_builder.py
@@ -394,7 +394,9 @@ def build_mtgjson_set(set_code: str) -> Optional[MtgjsonSetObject]:
     mtgjson_set.total_set_size = base_total_sizes[1]
 
     # Building cards is a process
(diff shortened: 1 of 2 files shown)


mtgjson5/set_builder.py
```diff
@@ -683,6 +685,10 @@ def build_mtgjson_card(
         if card_type not in {"starterdeck", "planeswalkerdeck"}
     ]
 
+    card_release_date = scryfall_object.get("released_at")
+    if set_release_date != "" and set_release_date != card_release_date:
```
ZeldaZach: This can be shortened to 
```suggestion
    if set_release_date and set_release_date != card_release_date:
```

ZeldaZach:
Please make sure this passes tox. You can run the toxfile locally and see the outputs.

kodawah:
updated as requested

Issue #689, opened 2020-09-23
Runtime Warning from incompatible Gevent version


0az:
#### Version(s) affected
- [ ] MTGJSON 4
- [X] MTGJSON 5

#### Version(s) tested against
5.0.1-4-g64e4e0a

#### Description of Bug

mtgjson5 exits with SIGSEV when started with Python 3.8.
```
<frozen importlib._bootstrap>:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
```

```dockerfile
FROM python:3.8.5-buster

WORKDIR /app

ADD \
	requirements.txt \
	/app/
RUN pip install -r requirements.txt

ADD \
	MANIFEST.in \
	README.md \
	setup.py \
	/app/
ADD \
	mtgjson5/ \
	/app/mtgjson5
RUN pip install -e .

ENTRYPOINT ["python", "-m", "mtgjson5"]
```

0az:
See: https://github.com/python-greenlet/greenlet/issues/178

Fixing this also means dropping support for 3.6.

jamadden:
Why would this mean dropping support for Python 3.6? gevent 20.9 supports the same versions of Python as gevent 20.6.x did, and the same goes for greenlet 0.4.17 and 0.4.16.

ZeldaZach:
My goal is to support py 3.6 still, which is why I'm not overly happy with the fix. Will do some testing on the latest version in a few days to see what happens on 3.6

ZeldaZach:
Should be addressed now

PR #690, opened 2020-09-23
Fix sigsev on Python 3.8 (#689)

0az:
This also drops support for Python 3.6, as the latest versions of Gevent/Greenlet upstreams are not binary compatible with 3.6.

The other solution is to pin Greenlet's version, but that's just kicking the entire issue down the road.

[pep-0494]: https://www.python.org/dev/peps/pep-0494


ZeldaZach:
Based on comments in #689, this isn't the fix I'd like to go with. Dropping Py3.6 doesn't seem to be necessary at this point. Will look into pinning greenlet instead

PR #692, opened 2020-10-03
RFC - daily prices + detailed prices

kodawah:
I hacked together something to provide a detailed daily price file instead of the complete price history, this is incomplete but i'd like to get some early feedback

1. is it ok to produce two separate files and two calls for each providers every day?
2. is it ok to change the output of alldailyprices?
3. how can we populate custom tags from the provided to the final json? I tried adding an optional dictionary and do more hacks in ck, but not sure if that's the right approach

### DIFF
--- a/mtgjson5/__main__.py
+++ b/mtgjson5/__main__.py
@@ -18,6 +18,7 @@ from mtgjson5.consts import CONFIG_PATH, MTGJSON_VERSION, OUTPUT_PATH
 from mtgjson5.output_generator import (
     generate_compiled_output_files,
     generate_compiled_prices_output,
+    generate_compiled_daily_prices_output,
     generate_output_file_hashes,
     write_set_file,
 )
@@ -83,6 +84,7 @@ def dispatcher(args: argparse.Namespace) -> None:
     # If a price build, simply build prices and exit
     if args.price_build:
         generate_compiled_prices_output(build_prices(), args.pretty)
+        generate_compiled_daily_prices_output(build_prices(True), args.pretty)
         if args.compress:
             compress_mtgjson_contents(OUTPUT_PATH)
         generate_output_file_hashes(OUTPUT_PATH)
--- a/mtgjson5/classes/mtgjson_prices.py
+++ b/mtgjson5/classes/mtgjson_prices.py
@@ -17,6 +17,7 @@ class MtgjsonPricesObject:
     buy_foil: Optional[float]
     sell_normal: Optional[float]
     sell_foil: Optional[float]
+    detailed: Optional[Dict[str, Any]]
 
     def __init__(self, source: str, provider: str, date: str, currency: str) -> None:
         """
@@ -37,6 +38,13 @@ class MtgjsonPricesObject:
         :return: JSON serialized object
         """
         buy_sell_option: Dict[str, Any] = {}
+        buy_sell_option["currency"] = self.currency
+
+        if self.detailed is not None:
+            buy_sell_option["retail"] = self.detailed["retail"]
+            return_object: Dict[str, Any] = {self.source: {self.provider: buy_sell_option}}
+            return return_object
+
         if (self.buy_normal is not None) or (self.buy_foil is not None):
             buy_sell_option["buylist"] = {"normal": {}, "foil": {}}
             if self.buy_normal is not None:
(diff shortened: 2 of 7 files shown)


mtgjson5/__main__.py
```diff
@@ -83,6 +84,7 @@ def dispatcher(args: argparse.Namespace) -> None:
     # If a price build, simply build prices and exit
     if args.price_build:
         generate_compiled_prices_output(build_prices(), args.pretty)
+        generate_compiled_daily_prices_output(build_prices(True), args.pretty)
```
ZeldaZach, line 87: Always best to specify what the "true" means. arg_name=True

mtgjson5/classes/mtgjson_prices.py
```diff
@@ -17,6 +17,7 @@ class MtgjsonPricesObject:
     buy_foil: Optional[float]
     sell_normal: Optional[float]
     sell_foil: Optional[float]
+    detailed: Optional[Dict[str, Any]]
```
ZeldaZach, line 20: A Dict in itself can be empty, as such you don't need to make this optional

mtgjson5/classes/mtgjson_prices.py
```diff
@@ -37,6 +38,13 @@ def to_json(self) -> Dict[str, Any]:
         :return: JSON serialized object
         """
         buy_sell_option: Dict[str, Any] = {}
+        buy_sell_option["currency"] = self.currency
+
+        if self.detailed is not None:
```
ZeldaZach, line 43: `if self.detailed` is fine

mtgjson5/compiled_classes/mtgjson_structures.py
```diff
@@ -56,6 +57,7 @@ def __init__(self) -> None:
         self.all_printings = "AllPrintings"
         self.atomic_cards = "AtomicCards"
         self.all_prices = "AllPrices"
+        self.all_daily_prices = "AllDailyPrices"
```
ZeldaZach, line 60: AllPricesToday please

mtgjson5/providers/cardkingdom.py
```diff
@@ -81,9 +82,13 @@ def generate_today_price_dict(
 
             mtgjson_uuid = card_kingdom_id_to_mtgjson[card_id]
 
+            key = self.today_date
+            if detailed_mode:
```
ZeldaZach, line 86: I don't agree with this change. AllPricesToday should be the same structure as AllPrices, except it only has one entry per op. This makes it easily changeable between the AllPrices file.

ZeldaZach:
Overall, this is a great start. A few key things I'd like to (re)point out.

1) I'd like this file to be called `AllPricesToday` if possible
2) I'd like this file to have the same structure as `AllPrices`, except it only has (up to) one entry per `foil` or `normal` tag

Thanks for the help, it's much appreciated!!

kodawah:
ok on the code changes, now onto the main topic

> I don't agree with this change. AllPricesToday should be the same structure as AllPrices, except it only has one entry per op. This makes it easily changeable between the AllPrices file.
> I'd like this file to have the same structure as AllPrices, except it only has (up to) one entry per foil or normal tag

So I see the rationale but there are a couple of limitations with that approach
1. the date is already available in `meta` and it's always going to be the same, so it's a lot of repeated information
2. having a fixed format kinda limits us, in case there is a vendor that exports percentages or other data
3. if we only keep the same format, that means there will be a lot of seller entries. For cardkingdom it doesn't really matter, but tcg exports 4 prices, and cardmarket exports 6 i think

what do you think we should do?

ZeldaZach:
I'd like to have a way to make our current AllPrices file meet the demand of these formats, and then extend it to the AllPricesTody file. Can we come up with a backwards compatible way to add the extra pricing data to AllPrices?

kodawah:
My opinion on AllPrices is known, I think it should not exist at all in its current form, so I don't believe any work should be spent on making any backwards compatibility for it. Additionally one of the problems of AllPrices is that it's too big, so I don't think adding even more data is going to have a positive effect. 

AllPricesToday is a new file, unless there is a clear value to it, I'd suggest to not shackle it to an old design.

kodawah:
closing this for now, too many incomplete changes

Issue #624, opened 2020-07-14
uuid to mcmId far from always correct on special cards


billylarsson:
#### Version(s) affected
- [x] MTGJSON 4
- [x] MTGJSON 5

#### Description of Bug

copy/paste from discord:

i think you got some (those i checked) cards with Flavor names wrongly connected from cardmarkets prices in AllPrices.
example: 6f8ddb82-805d-573a-beca-5d8f369c8108

The human eye knows that uuid: 6f8ddb82-805d-573a-beca-5d8f369c8108 should be mcmId: 451838 and not 451828 (as currently inputed)

approach:

Firstly I download the cardmarket’s expansion and cards data. 
"GET", "https://api.cardmarket.com/ws/v2.0/output.json/games/1/expansions"
"GET", f"https://api.cardmarket.com/ws/v2.0/output.json{endLink}"
I store them locally for further and reuse and expand if something new comes up (its currently like 50mb or so for everything)

Then from that data I fill out the blanks in MTGJson sets, mcmId and mcmName. (There’s more blanks in V5 than V4).

Next in MTGJson cards I set  mcmId’s to “NULL” because many of them are wrong . (I only find its useful to choose the post-pioneer cards)

Then since everyone is null, I take the easy ones first.
if setCode == setCode and cardNumber == cardNumer: ; update MTGJson/cards/mcmId with the one from MCM’s databank.
Next I match cardNumber -s-ending and cardNumber -p-ending and use isFoilOnly == True to the MCM prereleasepromo, also those are sometimes separable from using release-date since the prereleasepromo are sometimes dated earlier.
After positioning say three cards with the same name with highest possible accuracy out of four candidates, if both MTGJson and MCM-databank has four of that card means that the only card remaining from both databases may very well be a good match, like playing Sudoku.

If one are to look at row 318 in AllPrintings/sets it is clear that “PIKO” hasn't any mcmId or mcmName, I would assume if I were to debug the issue with the knowledge of your code that I would start there, perhaps that's connected to why many of the Ikoria cards are mismatched or left without an mcmId or even sharing the same mcmId while still being different at both ends.
A guess why the setCodes/names are'nt matched can be because cardmarket use a different naming, thay have two : colons while "you" just have one.

`
if set_name.lower() in self.set_map.keys():
            return str(self.set_map[set_name.lower()]["mcmName"])
`

[...]

PR #700, opened 2020-10-10
Fix extended arts mcmIds

mjainta:
Hey,

This intends to improve the mcmId matching regarding extendedart.
MCM uses set names suffixed with ": Extras" to store them.
When adding the mcm_details I get a second set, the extras_set and match for each card if we should use the regular or the extras set.
I did a quick comparison using ELD and see only mcmIds and purchaseUrls have changed. Attaching the files here.
[pr_mtgjson_fix_extended_arts.zip](https://github.com/mtgjson/mtgjson/files/5359043/pr_mtgjson_fix_extended_arts.zip)
A before/after test of 10E showed both JSONs to be identical.

I just have seen that you have opened #699 
If that conflicts in any way and your change will fix the extendedarts too, that would be also great.

Question: I am not sure how to test every set, don't wanna run the whole build at once and only have mcm credentials available.
If you could assist in any way checking if that change does have any side effects that would be great.

### DIFF
--- a/mtgjson5/set_builder.py
+++ b/mtgjson5/set_builder.py
@@ -1041,7 +1041,7 @@ def add_mcm_details(mtgjson_set: MtgjsonSetObject) -> None:
     for mtgjson_card in mtgjson_set.cards:
         delete_key = False
 
-        if "extendedart" in mtgjson_card.frame_effects and extras_cards:
+        if "boosterfun" in mtgjson_card.promo_types and extras_cards:
             # It is an extended art, search in the "Extras" set instead
             search_cards = extras_cards
         else:
--- a/mtgjson5/set_builder.py
+++ b/mtgjson5/set_builder.py
@@ -1041,8 +1041,9 @@ def add_mcm_details(mtgjson_set: MtgjsonSetObject) -> None:
     for mtgjson_card in mtgjson_set.cards:
         delete_key = False
 
+        # "boosterfun" is an alias for frame_effects=showcase, frame_effects=extendedart, and border_color=borderless
         if "boosterfun" in mtgjson_card.promo_types and extras_cards:
-            # It is an extended art, search in the "Extras" set instead
+            # It is an extended art, showcase, or borderless, search in the "Extras" set instead
             search_cards = extras_cards
         else:
             search_cards = mkm_cards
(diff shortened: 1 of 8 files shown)


mtgjson5/set_builder.py
```diff
@@ -1008,26 +1009,36 @@ def add_mcm_details(mtgjson_set: MtgjsonSetObject) -> None:
...
     LOGGER.info(f"Adding MCM details for {mtgjson_set.code}")
     mkm_cards = CardMarketProvider().get_mkm_cards(mtgjson_set.mcm_id)
+
+    if mtgjson_set.mcm_id_extras:
+        extras_cards = CardMarketProvider().get_mkm_cards(mtgjson_set.mcm_id_extras)
+
     for mtgjson_card in mtgjson_set.cards:
         delete_key = False
 
+        if "extendedart" in mtgjson_card.frame_effects and extras_cards:
```
kodawah: i think you should only check for `promo_type=boosterfun` which is an alias for`frame_effects=showcase`, `frame_effects=extendedart`, and `border_color=borderless`

mjainta: Good catch, that solves more cases at once :+1: 
Edit: Changed it

kodawah: thanks, however I don't know how the search will behave, for example if there are multiple cards with the same name, except one is showcase and the other one is borderless. Maybe let's wait for @ZeldaZach to chime in

ZeldaZach: Hey folks, thanks for taking the time to bring this to my attention!

Yeah this was my only real concern here. I know upstream has started providing some basic IDs for MCM, which we're now taking advantage of. However, as you can see from my https://github.com/mtgjson/mtgjson/issues/624#issuecomment-699571753, there is a fundamental flaw on the ID system still. I _think_ this PR is unrelated to that issue, and is solely an extension of the service, which is why I'm fine with it as-is.

mjainta: With upstream you mean scryfall?

Yea I dont think this solves all issues with the mcmid but my goal was to fix the extended arts. Thanks to @kodabb it will fix more tho :+1:

mjainta:
Hey @ZeldaZach 
any chance of getting this merged in the next time?

ZeldaZach (approved):
Thanks so much for your work! I pushed a small tox change and cleanup :)

PR #772, opened 2021-03-04
Basic MTGJSON Sealed Product Implementation

tpblaster:
This is a basic implementation of how MTGJSON could add a sealed product class to the set data files. This new class heavily leans on the TCGPlayer API for data as there is already infrastructure within MTGJSON to get data from the TCGPlayer API based on the set. This data is then compiled into a new class called MtgjsonSealedProductObject which contains the name, UUID, identifiers, purchase URLs, and release date of sealed product stored on TCGPlayer. This data is currently very barebones and could greatly be expanded upon by adding a contents field and combing existing MTGJSON data about boosters and decks into the product data. I think the current data is solid but it definitely needs a way to parse raw purchase URLs to ones with MTGJSON referral codes as the current referral system leans on the card class. Overall I think this is a good starting point for the addition of sealed product data. I want to thank Koda for helping with some debugging and design ideas. I have also included the WAR set file that is created with this current fork. 

WAR MTGJSON File:
[WAR.zip](https://github.com/mtgjson/mtgjson/files/6079974/WAR.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,7 +123,7 @@ compiled_outputs/
 /trace.html
 /tests/test_*.yaml
 /test_results/*.xml
-mtgjson.properties
+mtgjson5/resources/mtgjson.properties
 *.sqlite
 set_translations.json
 stocks_data.json
--- a/.gitignore
+++ b/.gitignore
@@ -123,7 +123,7 @@ compiled_outputs/
 /trace.html
 /tests/test_*.yaml
 /test_results/*.xml
-mtgjson5/resources/mtgjson.properties
+mtgjson.properties
 *.sqlite
 set_translations.json
 stocks_data.json
--- a/mtgjson5/classes/mtgjson_sealed_product.py
+++ b/mtgjson5/classes/mtgjson_sealed_product.py
@@ -25,8 +25,6 @@ class MtgjsonSealedProductObject:
         self.purchase_urls = MtgjsonPurchaseUrlsObject()
         self.raw_purchase_urls = {}
 
-
-
     def to_json(self) -> Dict[str, Any]:
         """
         Support json.dump()
--- a/mtgjson5/classes/mtgjson_sealed_product.py
+++ b/mtgjson5/classes/mtgjson_sealed_product.py
@@ -1,0 +1,38 @@
+"""
+MTGJSON Singular Sealed Product Object
+"""
+from typing import Any, Dict
+
+from mtgjson5.utils import to_camel_case
+
+from .mtgjson_purchase_urls import MtgjsonPurchaseUrlsObject
+from .mtgjson_identifiers import MtgjsonIdentifiersObject
+
+
+class MtgjsonSealedProductObject:
+    """
+    MTGJSON Singular Sealed Product Object
+    """
+
+    name: str
+    uuid: str
+    identifiers: MtgjsonIdentifiersObject
+    purchase_urls: MtgjsonPurchaseUrlsObject
+    raw_purchase_urls: Dict[str, str]
+    release_date: str
+
+    def __init__(self) -> None:
+        self.purchase_urls = MtgjsonPurchaseUrlsObject()
+        self.raw_purchase_urls = {}
+
+    def to_json(self) -> Dict[str, Any]:
+        """
+        Support json.dump()
+        :return: JSON serialized object
+        """
+
+        return {
+            to_camel_case(key): value
+            for key, value in self.__dict__.items()
+            if "__" not in key and not callable(value)
+        }
--- a/mtgjson5/classes/mtgjson_sealed_product.py
+++ b/mtgjson5/classes/mtgjson_sealed_product.py
@@ -4,7 +4,6 @@ MTGJSON Singular Sealed Product Object
 from typing import Any, Dict
 
 from ..utils import to_camel_case
-
 from .mtgjson_identifiers import MtgjsonIdentifiersObject
 from .mtgjson_purchase_urls import MtgjsonPurchaseUrlsObject
 
--- a/mtgjson5/classes/mtgjson_sealed_product.py
+++ b/mtgjson5/classes/mtgjson_sealed_product.py
@@ -1,7 +1,7 @@
 """
 MTGJSON Singular Sealed Product Object
 """
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from ..utils import to_camel_case
 from .mtgjson_identifiers import MtgjsonIdentifiersObject
@@ -18,7 +18,10 @@ class MtgjsonSealedProductObject:
     identifiers: MtgjsonIdentifiersObject
     purchase_urls: MtgjsonPurchaseUrlsObject
     raw_purchase_urls: Dict[str, str]
-    release_date: str
+    release_date: Optional[str]
+    __skip_keys = [
+        "raw_purchase_urls",
+    ]
 
     def __init__(self) -> None:
         self.identifiers = MtgjsonIdentifiersObject()
@@ -30,9 +33,10 @@ class MtgjsonSealedProductObject:
         Support json.dump()
         :return: JSON serialized object
         """
+        skip_keys = self.__skip_keys
 
         return {
             to_camel_case(key): value
             for key, value in self.__dict__.items()
-            if "__" not in key and not callable(value)
+            if "__" not in key and not callable(value) and key not in skip_keys
         }
--- a/mtgjson5/classes/mtgjson_sealed_product.py
+++ b/mtgjson5/classes/mtgjson_sealed_product.py
@@ -21,9 +21,11 @@ class MtgjsonSealedProductObject:
     release_date: str
 
     def __init__(self) -> None:
+        self.identifiers = MtgjsonIdentifiersObject()
         self.purchase_urls = MtgjsonPurchaseUrlsObject()
         self.raw_purchase_urls = {}
-        self.identifiers = MtgjsonIdentifiersObject()
+
+
 
     def to_json(self) -> Dict[str, Any]:
         """
--- a/mtgjson5/classes/mtgjson_sealed_product.py
+++ b/mtgjson5/classes/mtgjson_sealed_product.py
@@ -5,8 +5,8 @@ from typing import Any, Dict
 
 from mtgjson5.utils import to_camel_case
 
-from .mtgjson_purchase_urls import MtgjsonPurchaseUrlsObject
 from .mtgjson_identifiers import MtgjsonIdentifiersObject
+from .mtgjson_purchase_urls import MtgjsonPurchaseUrlsObject
 
 
 class MtgjsonSealedProductObject:
--- a/mtgjson5/classes/mtgjson_sealed_product.py
+++ b/mtgjson5/classes/mtgjson_sealed_product.py
@@ -18,7 +18,6 @@ class MtgjsonSealedProductObject:
     identifiers: MtgjsonIdentifiersObject
     purchase_urls: MtgjsonPurchaseUrlsObject
     raw_purchase_urls: Dict[str, str]
-    release_date: str
 
     def __init__(self) -> None:
         self.identifiers = MtgjsonIdentifiersObject()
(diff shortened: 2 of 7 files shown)


mtgjson5/providers/tcgplayer.py
```diff
@@ -250,6 +281,43 @@ def get_tcgplayer_sku_data(group_id_and_name: Tuple[str, str]) -> List[Dict[str,
...
+    api_offset = 0
+
+    while True:
+        api_response = TCGPlayerProvider().download(
+            "https://api.tcgplayer.com/catalog/products",
+            {
+                "offset": str(api_offset),
+                "limit": 100,
+                "categoryId": 1,
+                "getExtendedFields": True,
```
kodawah: what is this needed for?

mtgjson5/providers/tcgplayer.py
```diff
@@ -250,6 +281,43 @@ def get_tcgplayer_sku_data(group_id_and_name: Tuple[str, str]) -> List[Dict[str,
...
+            },
+        )
+
+        if not api_response:
+            # No more entries
+            break
+
+        response = json.loads(api_response)
+        if not response["results"]:
+            # Something went wrong
```
kodawah: maybe return the error array somewhere?

mtgjson5/set_builder.py
```diff
@@ -412,6 +413,12 @@ def build_mtgjson_set(set_code: str) -> Optional[MtgjsonSetObject]:
     mtgjson_set.tcgplayer_group_id = set_data.get("tcgplayer_id")
     mtgjson_set.booster = GitHubBoostersProvider().get_set_booster_data(set_code)
 
+    mtgjson_set.sealed_product = (
+        TCGPlayerProvider().generate_mtgjson_sealed_product_objects(
+            mtgjson_set.tcgplayer_group_id
```
kodawah, line 420: this will ping tcg api for each set, is that something ok to do? or should it be downloading the entirety of sealed stuff separately and then group them later?

kodawah: also some sets do not have tcgplayer_group_id, you can skip those i believe

mtgjson5/providers/tcgplayer.py
tpblaster: It is unnecessary for now but could be used to add a content description field that TCG stores

kodawah: let's add it when needed then? the returned data seems either missing or useless, like

```json
            "extendedData": [
                {
                    "name": "OracleText",
                    "displayName": "Rules Text Contains",
                    "value": "Booster box for 10th Edition."
                }
            ]
```

mtgjson5/classes/mtgjson_sealed_product.py
```diff
@@ -0,0 +1,39 @@
+"""
+MTGJSON Singular Sealed Product Object
+"""
+from typing import Any, Dict
+
+from mtgjson5.utils import to_camel_case
```
ZeldaZach: Import should be relative

mtgjson5/classes/mtgjson_set.py
```diff
@@ -34,6 +35,7 @@ class MtgjsonSetObject:
     parent_code: str
     release_date: str
     tcgplayer_group_id: Optional[int]
+    sealed_product: List[MtgjsonSealedProductObject]
```
ZeldaZach, line 38: Do we like this name?

mtgjson5/providers/tcgplayer.py
```diff
@@ -213,6 +215,35 @@ def generate_today_price_dict(
...
+        if not self.__keys_found:
+            LOGGER.warning("Keys not found for TCGPlayer, skipping")
+            return []
+
+        sealed_data = get_tcgplayer_sealed_data(group_id)
+
+        mtgjson_sealed_products = []
+
+        for product in sealed_data:
+            sealed_product: MtgjsonSealedProductObject = MtgjsonSealedProductObject()
```
ZeldaZach: Not sure you need to define the type here

mtgjson5/providers/tcgplayer.py
```diff
@@ -213,6 +215,35 @@ def generate_today_price_dict(
...
+            sealed_product.name = product["cleanName"]
+            sealed_product.raw_purchase_urls["tcgplayer"] = product["url"]
+            sealed_product.identifiers.tcgplayer_product_id = product["productId"]
+            sealed_product.uuid = str(
+                uuid.uuid5(uuid.NAMESPACE_DNS, sealed_product.name)
```
ZeldaZach: Are names guaranteed not to change?

mtgjson5/providers/tcgplayer.py
ZeldaZach: I agree with not adding un-used code

mtgjson5/providers/tcgplayer.py
ZeldaZach: Should log an error

mtgjson5/set_builder.py
ZeldaZach: If sealed stuff can be downloaded at once, that might be better, re; Koda's info

ZeldaZach (changes requested):

mtgjson5/set_builder.py
tpblaster: It is much harder and doesn't map well to how sets are currently built, also doesn't save many requests,

kodawah:
how about a few more flags? mostly to know whether if the product is a booster with random cards or if it's a box set with a known card list (for ftv, commander sets and so on)

.gitignore
```diff
@@ -123,7 +123,7 @@ compiled_outputs/
 /trace.html
 /tests/test_*.yaml
 /test_results/*.xml
-mtgjson.properties
+mtgjson5/resources/mtgjson.properties
```
ZeldaZach: This should be rolled back b/c it should block any file named mtgjson.properties within the repo

mtgjson5/classes/mtgjson_sealed_product.py
```diff
@@ -0,0 +1,38 @@
...
+
+    name: str
+    uuid: str
+    identifiers: MtgjsonIdentifiersObject
+    purchase_urls: MtgjsonPurchaseUrlsObject
+    raw_purchase_urls: Dict[str, str]
+    release_date: str
+
+    def __init__(self) -> None:
+        self.purchase_urls = MtgjsonPurchaseUrlsObject()
```
ZeldaZach, line 28: nit: These should be initialized in order to the variables above

mtgjson5/referral_builder.py
```diff
@@ -44,6 +44,22 @@ def build_referral_map(mtgjson_set: MtgjsonSetObject) -> List[Tuple[str, str]]:
                     ),
                 )
             )
+    for mtgjson_sealed_object in mtgjson_set.sealed_product:
```
ZeldaZach, line 34: Could you take the above method you copied, functionize it, and then call that function twice instead?

mtgjson5/providers/tcgplayer.py
```diff
@@ -213,6 +215,42 @@ def generate_today_price_dict(
...
+
+        for product in sealed_data:
+            sealed_product = MtgjsonSealedProductObject()
+            sealed_product.name = product["cleanName"]
+            sealed_product.identifiers.tcgplayer_product_id = str(product["productId"])
+            sealed_product.uuid = str(
+                uuid.uuid5(uuid.NAMESPACE_DNS, sealed_product.name)
+            )
+            sealed_product.release_date = product["presaleInfo"]["releasedOn"]
+            sealed_product.raw_purchase_urls[
```
ZeldaZach, line 239: remove

mtgjson5/providers/tcgplayer.py
```diff
@@ -213,6 +215,42 @@ def generate_today_price_dict(
...
+        mtgjson_sealed_products = []
+
+        for product in sealed_data:
+            sealed_product = MtgjsonSealedProductObject()
+            sealed_product.name = product["cleanName"]
+            sealed_product.identifiers.tcgplayer_product_id = str(product["productId"])
+            sealed_product.uuid = str(
+                uuid.uuid5(uuid.NAMESPACE_DNS, sealed_product.name)
+            )
+            sealed_product.release_date = product["presaleInfo"]["releasedOn"]
```
ZeldaZach: This is nullable in its current form. Is that correct?

ZeldaZach: Also, can you convert it to `YYYY-MM-DD` w/o timestamps, as they're mostly pointless

mtgjson5/providers/tcgplayer.py
```diff
@@ -213,6 +215,42 @@ def generate_today_price_dict(
...
+
+        sealed_data = get_tcgplayer_sealed_data(group_id)
+
+        mtgjson_sealed_products = []
+
+        for product in sealed_data:
+            sealed_product = MtgjsonSealedProductObject()
+            sealed_product.name = product["cleanName"]
+            sealed_product.identifiers.tcgplayer_product_id = str(product["productId"])
+            sealed_product.uuid = str(
```
ZeldaZach: Would prefer this to be added to the add_uuid function already defined in set_builder. Maybe move that function to utils and change it. I just want all UUID stuff in one location

ZeldaZach (changes requested):

mtgjson5/classes/mtgjson_sealed_product.py
```diff
@@ -0,0 +1,42 @@
...
+        self.identifiers = MtgjsonIdentifiersObject()
+        self.purchase_urls = MtgjsonPurchaseUrlsObject()
+        self.raw_purchase_urls = {}
+
+    def to_json(self) -> Dict[str, Any]:
+        """
+        Support json.dump()
+        :return: JSON serialized object
+        """
+        skip_keys = self.__skip_keys
```
ZeldaZach: Just remove this line and on 41 make it "not in self.__skip_keys" since it's a const

mtgjson5/providers/tcgplayer.py
```diff
@@ -11,6 +11,7 @@
 from singleton_decorator import singleton
 
 from ..classes import MtgjsonPricesObject
+from ..classes.mtgjson_sealed_product import MtgjsonSealedProductObject
```
ZeldaZach: You'll want to update the provider/__init__ file to import sealed product.

mtgjson5/providers/tcgplayer.py
```diff
@@ -213,6 +214,38 @@ def generate_today_price_dict(
 
         return dict(combined_listings)
 
+    def generate_mtgjson_sealed_product_objects(
+        self, group_id: Optional[Any]
```
ZeldaZach: group_id is scary with that type. Make it more restrictive

mtgjson5/providers/tcgplayer.py
```diff
@@ -250,6 +283,43 @@ def get_tcgplayer_sku_data(group_id_and_name: Tuple[str, str]) -> List[Dict[str,
     return magic_set_product_data
 
 
+def get_tcgplayer_sealed_data(group_id: Optional[Any]) -> List[Dict[str, Any]]:
```
ZeldaZach: same here, group_id needs to be more restrictive

mtgjson5/set_builder.py
```diff
@@ -20,6 +20,7 @@
     MtgjsonRulingObject,
     MtgjsonSetObject,
 )
+from .classes.mtgjson_sealed_product import MtgjsonSealedProductObject
```
ZeldaZach: should be imported to classes/__init__.py instead

mtgjson5/set_builder.py
```diff
@@ -433,6 +444,27 @@ def build_base_mtgjson_tokens(
...
+        add_uuid(sealed_product)
+
+
+def add_sealed_purchase_url(mtgjson_set: MtgjsonSetObject) -> None:
+    """
+    Adds all purhcase urls to each sealed product object within a set
+    :param mtgjson_set: the set to add purchase urls to
+    """
+    for sealed_product in mtgjson_set.sealed_product:
+        if sealed_product.identifiers.tcgplayer_product_id is not None:
```
ZeldaZach: "is not None" isn't required, since we're just doing a simple empty / null check

ZeldaZach (approved):
Fails 3.6, will look for workaround at some point. We are building 3.8 afaik.

kodawah:
🎉 🎉 🎉 🎉

PR #796, opened 2021-05-02
Add even more sealed product types

kodawah:
Don't we love undocumented product types?
/cc @tpblaster


tpblaster:
I just think that my approach towards finding them was wrong I guess. I used the website search and maybe I picked a set without fat packs. I think they use their own api for the website search so I must imagine that is the best way to find any missing products.

kodawah:
> I just think that my approach towards finding them was wrong I guess.

Your approach was fine, these fields are totally undocumented in tcg api (and it's insane there are 13 terms to specify "not-cards")

PR #833, opened 2021-08-30
Handle Gatherer crapping out on a few Multiverse IDs

ZeldaZach:
by fixing the retry logic used internally. Also handles the crashes that happen when Gatherer goes down.
 I catch the three exception types I've seen from gatherer depending on if a card isn't found or if gatherer redirects and is completely down


kodawah:
thanks for the quick fix!

PR #852, opened 2021-09-18
ADD "finishes" for TCGPlayerSku file

### LINKED ISSUE #847: Update TcgplayerSkus with finish information
https://mtgjson.com/abstract-models/tcgplayer-skus/

For sku id, we need to add which finish among the possible values.

I propose

```json
{
  "<Card UUID>": [
    {
      "finish": "<String>",
      "condition": "<String>",
      "language": "<String>",
      "printing": "<String>",
      "productId": <Int>,
      "skuId": <Int>
    },
    ... // More SKU objects
  ],
  ... // More card UUIDs
}
```

but I would not be against a more complex structure like

```json
{
  "<Card UUID>": {
    "nonfoil" : [
      {
      "condition": "<String>",
      "language": "<String>",
      "printing": "<String>",
      "productId": <Int>,
      "skuId": <Int>
      },
      ... // More SKU objects with the same finish
    ],
  },
  "foil" : { /* Same as above */ },
  "etched" : { /* Same as above */ }
  },
  ... // More card UUIDs
}
```

@tpblaster any opinions?

ZeldaZach:
giving a great view into what the card's finish is. Fix #847

ADD
TCGPlayerSku entry finish

### DIFF
--- a/mtgjson5/providers/tcgplayer.py
+++ b/mtgjson5/providers/tcgplayer.py
@@ -24,12 +24,7 @@ class CardFinish(enum.Enum):
     Self Driven by MTGJSON
     """
 
-    ALTERNATE_ART = "Alternate Art"
-    BORDERLESS = "Borderless"
-    EXTENDED_ART = "Extended Art"
     FOIL_ETCHED = "Foil Etched"
-    GLOSSY = "SDCC 2019 Exclusive"
-    RETRO = "Retro Frame"
 
     @classmethod
     def has_value(cls, value: str) -> bool:
@@ -508,7 +503,6 @@ def get_card_finish(card_name: str) -> Optional[str]:
     card_finishes = re.findall(r"\(([^)0-9]+)\)", card_name)
     for card_finish in card_finishes:
         if not CardFinish.has_value(card_finish):
-            LOGGER.warning(f"Unknown TCGPlayer card finish: {card_finish}")
             continue
 
         result_card_finish = CardFinish(card_finish).name.replace("_", " ")
--- a/mtgjson5/providers/tcgplayer.py
+++ b/mtgjson5/providers/tcgplayer.py
@@ -5,6 +5,7 @@ import enum
 import json
 import logging
 import pathlib
+import re
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import requests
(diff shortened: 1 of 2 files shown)


mtgjson5/providers/tcgplayer.py
```diff
@@ -17,6 +18,28 @@
...
+    TCGPlayer Card Condition
+    Self Driven by MTGJSON
+    """
+
+    ALTERNATE_ART = "Alternate Art"
+    BORDERLESS = "Borderless"
+    EXTENDED_ART = "Extended Art"
+    FOIL_ETCHED = "Foil Etched"
+    GLOSSY = "SDCC 2019 Exclusive"
+    RETRO = "Retro Frame"
```
kodawah: i don't get why all these variants are needed, especially because they are not technically "finishes", but different frame effects/borders/frame version

I think you could just check if the name contains "foil etched" and mark it as such, no?

mtgjson5/providers/tcgplayer.py
```diff
@@ -471,16 +494,51 @@ def get_tcgplayer_prices_map(
...
+    for sku in skus:
+        entry = {
+            "skuId": sku["skuId"],
+            "productId": sku["productId"],
+            "language": CardLanguage(sku["languageId"]).name.replace("_", " "),
+            "printing": CardPrinting(sku["printingId"]).name.replace("_", " "),
+            "condition": CardCondition(sku["conditionId"]).name.replace("_", " "),
+        }
+        if card_finish:
+            entry["finish"] = card_finish
```
kodawah, line 535: maybe make it all caps like the other fields?

mtgjson5/providers/tcgplayer.py
ZeldaZach: Is Foil Etched the only entity we want to support? I'd argue Glossy should be supported too. I understand why you might be against retro/extended, so i'll ask a few others about it

mtgjson5/providers/tcgplayer.py
ZeldaZach: It is

mtgjson5/providers/tcgplayer.py
kodawah: yes, foil etched is the only one entity we need. Glossy is tame so far as there are no duplicates finishes, so there is no need to duplicate them (and I'd argue we shouldn't unless they became a real thing from wotc)

note that I am not "against retro/extended", they simply are not finishes, each of those will have separate uuids, so there is no need to exposing them in this file like this

PR #868, opened 2021-11-14
ADD faceFlavorName, which is similar to faceName 

### LINKED ISSUE #865: VOW DFCs do not have FlavorName field
The new dual faced cards from VOW of the Dracula series do not have FlavorName set to the alternative name.

example `Voldaren Bloodcaster // Bloodbat Summoner` #.338 would need `Dracula, Lord of Blood // Dracula, Lord of Bats` as FlavorName

Note that I believe we should report the full alternative name in FlavorName, if we want to track different flavored face names we should introduce a new field like `FaceFlavorName` (in the example above "Dracula, Lord of Blood")

ZeldaZach:
but if there's a flavor name on the face. Seems simple enough. Also handles Dracula series cards not getting the flavorName attribute (and those will now also have faceFlavorName!)
[VOW.json.txt](https://github.com/mtgjson/mtgjson/files/7534460/VOW.json.txt)

Fix #865

### DIFF
--- a/mtgjson5/classes/mtgjson_card.py
+++ b/mtgjson5/classes/mtgjson_card.py
@@ -33,6 +33,7 @@ class MtgjsonCardObject:
     duel_deck: Optional[str]
     edhrec_rank: Optional[int]
     face_converted_mana_cost: float
+    face_flavor_name: Optional[str]
     face_mana_value: float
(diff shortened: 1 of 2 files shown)


kodawah:
👍

PR #891, opened 2021-12-17
Support Alchemy Cards Better (First Idea)

### LINKED ISSUE #885: Support for Alchemy Cards
This is a fun one. Baseline idea is we'll have duplicate entries for sets with "Rebalanced" cards, and establish a trait so we can easily filter them in/out when necessary. Rebalanced cards will find their home in the original set the card was printed/modified from (at least for now). They will have the alternative flag established to indicate it's different, as well as a different collector number.

Since any entity of a card can be changed (apart from the name... _please_), rebalanced cards will be a bit of a pain with UUID generation since we seed on that component. Might have to change how we generate these cards UUIDs to be based on a more stable source to avoid constant re-hashing.

For AtomicCards, there just needs to be N entries at the key level -- "CardName" and "A-CardName" to allow for cohesion between the products.

ZeldaZach:
Fix #885 

This is a review-required PR to better support Alchemy cards. To start, I've added isRebalanced(bool) and originalPrintings(list[str]) to the card object.

For cards that have been rebalanced by Wizards, a new card entry will be put into the original set file that the card appeared in. This new card entry will have isRebalanced=True, isAlternative=True, and originalPrintings containing the UUID(s) of the card(s) in the same set file that were the "original" version. Alchemy cards will _not_ change the baseSetSize or totalSetSize of those sets, just simply add new values in.

Alchemy cards will be handled in AtomicCards as a new entry, which is fine. This doesn't handle the changing UUID of the card if they get re-re-balanced, so I'm scouting ideas on that.

My implementation files:
[KHM.json.txt](https://github.com/mtgjson/mtgjson/files/7733708/KHM.json.txt)
[diff.txt](https://github.com/mtgjson/mtgjson/files/7733707/diff.txt)

By setting isAlternative, we hopefully keep backwards compatibility that these cards are extra and can possibly be ignored. isRebalanced adds new struct giving users the ability to filter/use as they please.

This is just one idea I had on handling Alchemy better, but I'm open to alternative suggestions.

One alternative design I scouted was adding a new field to cards "rebalancedEntities(Dict[str, Any])" that contains a dict of all the keys changed and their new values. I like the concept, but I'd like to scout a few more ideas / feedback before committing to one.

### DIFF
--- a/mtgjson5/classes/mtgjson_card.py
+++ b/mtgjson5/classes/mtgjson_card.py
@@ -55,6 +55,7 @@ class MtgjsonCardObject:
     is_online_only: Optional[bool]
     is_oversized: Optional[bool]
     is_promo: Optional[bool]
+    is_rebalanced: Optional[bool]
     is_reprint: Optional[bool]
     is_reserved: Optional[bool]
     is_starter: Optional[bool]
@@ -71,6 +72,7 @@ class MtgjsonCardObject:
     mana_value: float
     name: str
     number: str
+    original_printings: List[str]
     original_release_date: Optional[str]
     original_text: Optional[str]
     original_type: Optional[str]
--- a/mtgjson5/classes/mtgjson_card.py
+++ b/mtgjson5/classes/mtgjson_card.py
@@ -83,6 +83,7 @@ class MtgjsonCardObject:
     promo_types: List[str]
     purchase_urls: MtgjsonPurchaseUrlsObject
     rarity: str
+    rebalanced_printings: List[str]
     reverse_related: Optional[List[str]]
     rulings: List[MtgjsonRulingObject]
     security_stamp: Optional[str]
--- a/mtgjson5/set_builder.py
+++ b/mtgjson5/set_builder.py
@@ -293,6 +293,28 @@ def parse_rulings(rulings_url: str) -> List[MtgjsonRulingObject]:
     return sorted(mtgjson_rules, key=lambda ruling: ruling.date)
 
 
+def add_rebalanced_to_original_linkage(mtgjson_set: MtgjsonSetObject) -> None:
+    """
+    When Wizards rebalances a card, they break the link between
+    the new card and the original card. We will create a one-way
+    linkage back to the original card, should that prove useful
+    to the end user.
+    :param mtgjson_set MTGJSON Set object
+    """
+    LOGGER.info(f"Linking rebalanced cards for {mtgjson_set.code}")
+
+    for card in mtgjson_set.cards:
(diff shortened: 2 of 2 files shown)


ZeldaZach:
cc: @kodabb

kodawah:
I'd favor not reusing `isAlternative` because it's something actively used for paper editions and I'd rather not overload the meaning with something online-only. Also I'd keep the rebalanced cards out of the main Cards array, again for backward compatibility, and to let applications that do not care about online cards avoid filtering explicitly for this.

I like the idea of a separate entity containings the cards that got rebalanced, with a full representation of the new object, so that applications can just replace the old entity with the new one rather than finding what exactly changed. I don't think it would be worthwhile to keep a per-object array to track history of the changes, but that may be considered too.

Depending on the implementation of this new entity, you might even skip adding `isRebalanced` to card objects, for example if you use a dictionary, as applications could just check the the card status with a single operation, but maybe using an array of the entity is more in-line with mtgjson representation.

AEFeinstein:
I like `isRebalanced`. At least it's better than my current hack, which is to check the name for `^A-.*`.

Do the rebalanced cards have [`isOnlineOnly`](https://mtgjson.com/data-models/card-set/#isonlineonly) set to true? The documentation specifies Magic Online, not Arena. Is that a misnomer?

Edit: yeah, it's set for cards like "A-Goldspan Dragon." The documentation should be updated to make that clearer.

staghouse:
> I like `isRebalanced`. At least it's better than my current hack, which is to check the name for `^A-.*`.
> 
> Do the rebalanced cards have [`isOnlineOnly`](https://mtgjson.com/data-models/card-set/#isonlineonly) set to true? The documentation specifies Magic Online, not Arena. Is that a misnomer?
> 
> Edit: yeah, it's set for cards like "A-Goldspan Dragon." The documentation should be updated to make that clearer.

Can you define which part of the documentation needs more clarity? I want to make sure to implement it properly.

AEFeinstein:
@staghouse, the documentation states:

> **isOnlineOnly**
If the card is only available in Magic: The Gathering Online.

The clarity is that Magic: The Gathering Online is a separate game than Magic: The Gathering Arena, and this field is being used for both the online games.

staghouse:
> @staghouse, the documentation states:
> 
> > **isOnlineOnly**
> > If the card is only available in Magic: The Gathering Online.
> 
> The clarity is that Magic: The Gathering Online is a separate game than Magic: The Gathering Arena, and this field is being used for both the online games.

Will be clarified in v5.2.1 documentation release

PR #902, opened 2021-12-30
Add GHA to build Dockerfile

ZeldaZach:
Use envvars over flags

Bump versions to support Win11, Python3.6-3.10

 Split off how constants are defined by a true constants (one that can't be modified) and config (one that is dependent upon a config file). Configs can now be loaded via SSM (AWS Parameter Store) or Localhost (resources/mtgjson.properties). Flag added to support this option. Also added S3 upload support for final output. Docker file ran locally and passed all params (Build, Prices, Sqlite, Compression, Upload). Fixup requirements & Setup


mtgjson5/__init__.py
```diff
@@ -1,5 +1,5 @@
 """
 MTGJSON, written by Zachary Halpern
 https://mtgjson.com/
-MIT License (c) 2019-2020
+MIT License (c) 2019-2022
```
kodawah, line 4: unrelated change

mtgjson5/__main__.py
```diff
@@ -36,9 +22,17 @@ def build_mtgjson_sets(
     """
     Build each set one-by-one and output them to a file
     :param sets_to_build: Sets to construct
-    :param output_pretty: Should we dump minified?
-    :param include_referrals: Should we include referrals?
+    :param output_pretty: Should we dump minified
+    :param include_referrals: Should we include referrals
```
kodawah, line 26: id leave the ? at the end (or at least drop them in a separate commit)

mtgjson5/arg_parser.py
```diff
@@ -22,51 +17,57 @@ def parse_args() -> argparse.Namespace:
...
+        action="store_true",
+        help="Use environment variables over parser flags for build operations",
+    )
+
     # What set(s) to build
     sets_group = parser.add_mutually_exclusive_group()
     sets_group.add_argument(
-        "-s",
         "--sets",
+        "-s",
```
kodawah, line 30: why?

ZeldaZach: By reordering, I changed how the help menu displays to show the full command instead of the shorthand. This is a non-breaking change and aims to upgrade how we help our users

mtgjson5/__main__.py
ZeldaZach: Moved to a commit

mtgjson5/__init__.py
ZeldaZach: Moved to a commit

PR #908, opened 2022-01-20
Sealed Decklist Basic Draft

tpblaster:
This is a basic implementation of the code to add decklists to sealed products. This is currently a draft and has a lot of issues.

### DIFF
--- a/mtgjson5/set_builder.py
+++ b/mtgjson5/set_builder.py
@@ -1143,8 +1143,10 @@ def add_variations_and_alternative_fields(mtgjson_set: MtgjsonSetObject) -> None
         # otherwise, it's an alternative printing
         distinct_card_printing = (
             f"{this_card.name}|{this_card.border_color}|{this_card.frame_version}|"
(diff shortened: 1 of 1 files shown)


mtgjson5/providers/tcgplayer.py
```diff
@@ -251,7 +251,7 @@ def generate_mtgjson_sealed_product_objects(
 
         for product in sealed_data:
             sealed_product = MtgjsonSealedProductObject()
-            sealed_product.name = product["cleanName"]
+            sealed_product.name = product["name"]
```
kodawah: why is this needed?

kodawah:
could you post an example output from this code?

tpblaster:
Here is the output for War of the Spark
[WAR.zip](https://github.com/mtgjson/mtgjson/files/7919357/WAR.zip)

kodawah:
@tpblaster I was hoping to check the output of a decklist-based set, like a commander edition which has multiple products containing different cards from the same set

PR #943, opened 2022-05-18
Support MultiverseBridge (AKA CardSphere) IDs & Prices

ZeldaZach:
Also abstract away part of how prices are generated for providers, since it's a lot of duplicated code.
(Will work more towards converting other price builders to use the generic method if possible.)

ADD card.identifiers.cardsphere_id -- Optional String
ADD set.cardsphere_set_id -- Optional Integer
ADD CardSphere to prices file under paper category with USD -- Retail only, No Buylist

### DIFF
--- a/mtgjson5/classes/mtgjson_identifiers.py
+++ b/mtgjson5/classes/mtgjson_identifiers.py
@@ -14,9 +14,11 @@ class MtgjsonIdentifiersObject:
     card_kingdom_etched_id: Optional[str]
     card_kingdom_foil_id: Optional[str]
     card_kingdom_id: Optional[str]
+    cardsphere_id: Optional[str]
     mcm_id: Optional[str]
     mcm_meta_id: Optional[str]
     mtg_arena_id: Optional[str]
+    mtgjson_v4_id: Optional[str]
     mtgo_foil_id: Optional[str]
     mtgo_id: Optional[str]
     multiverse_id: Optional[str]
@@ -25,7 +27,6 @@ class MtgjsonIdentifiersObject:
     scryfall_oracle_id: Optional[str]
     tcgplayer_etched_product_id: Optional[str]
     tcgplayer_product_id: Optional[str]
-    mtgjson_v4_id: Optional[str]
 
     def __init__(self) -> None:
         """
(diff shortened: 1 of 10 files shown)


mtgjson5/price_builder.py
```diff
@@ -74,9 +75,12 @@ def build_today_prices() -> Dict[str, Any]:
     tcgplayer = _generate_prices(TCGPlayerProvider())
     card_market = _generate_prices(CardMarketProvider())
     card_kingdom = _generate_prices(CardKingdomProvider())
+    cardsphere = _generate_prices(MultiverseBridgeProvider())
```
noob247365, line 78: Naming is not consistent. Should either rename provider class, or rename variable to match

noob247365: After further review, would side with renaming the variable over renaming the class

noob247365 (changes requested):
Just one variable name that's inconsistent, rest seems fine (for as much as I understand about this project :P)

mtgjson5/price_builder.py
ZeldaZach: The naming is intentionally different, as MB can/will provide other price Ids in the future, and this sets a precedent that we can pull it in multiple ways. If my logic here is wrong, I'm fine with changing it

noob247365: Oh, makes enough sense. Then yeah, keeping it is fine 👍

mtgjson5/classes/mtgjson_identifiers.py
```diff
@@ -14,9 +14,11 @@ class MtgjsonIdentifiersObject:
     card_kingdom_etched_id: Optional[str]
     card_kingdom_foil_id: Optional[str]
     card_kingdom_id: Optional[str]
+    cardsphere_id: Optional[str]
     mcm_id: Optional[str]
     mcm_meta_id: Optional[str]
     mtg_arena_id: Optional[str]
+    mtgjson_v4_id: Optional[str]
```
kodawah, line 21: unneeded move

kodawah (approved):

PR #946, opened 2022-05-27
ADD AllPricesToday as a smaller alternative

### LINKED ISSUE #673: smaller allprices.json file for daily price data
instead of downloading a huge (maybe compressed) file containing several days worth of data, it might be interesting to prive an abridged version containing only today's prices

while an xdelta will cover that in the future a lot of sites tracking prices have their own way of managing history, and do not need the backlog, saving gigabites of downloads over time, and using more than half less memory while operating on the file

### LINKED ISSUE #750: Design for AllPricesV2.json
This proposal describes the creation of a new file, supplanting most of the functionality of the current AllPrices.json and expanding the number of data points. Among the limitations of the current solution there is the fact that the generated file is pretty big in terms of size, can only host one type of price and it providing redundant information.

**Proposal**
The current nested dictionary structure works on a top level view, but a few fields should be moved elsewhere since they are constant. The main element is represented as a dictionary containing a foil and nonfoil price (assuming the card can be both), or just one of them (depending on the foil properties).

* *cardId*: is the mtgjson hash of the card
* *stroname*: is the name of the source providing the data points
* *price_type*: is the type of price represented, for example "buylist" "low" "market" "retail"

```json
"<cardId>": {
    "<storename>": {
        "<price_type>": {
             "foil": 0.0,
             "nonfoil": 0.0,
        },
        ...
    }
}
```

As noted this design offers for expansion in case new price type are added per store, or new stores are added.
The static properties such as `currency` and `whether it's paper or online prices` are per-store properties and should not be in the main data section.

```json
"static": {
    "<storename>": {
        "printing_type": "paper",
        "currency": "EUR"
    },
    ...
}
```

**Open questions**
- is it ok to break compatibility with the current design? hopefully the added benefits justify it
- where to add the static data? `meta` could be a possibility but a new section could be added too
- should this be worked on for 5.1 or later?

ZeldaZach:
- Gives support so MTGGraphQL can pull prices in
- Fix #673

[AllPricesToday.json.txt.zip](https://github.com/mtgjson/mtgjson/files/8789888/AllPricesToday.json.txt.zip)

### DIFF
--- a/mtgjson5/__main__.py
+++ b/mtgjson5/__main__.py
@@ -82,16 +82,15 @@ def dispatcher(args: argparse.Namespace) -> None:
     from mtgjson5.mtgjson_config import MtgjsonConfig
     from mtgjson5.mtgjson_s3_handler import MtgjsonS3Handler
     from mtgjson5.output_generator import (
+        build_price_files,
         generate_compiled_output_files,
-        generate_compiled_prices_output,
         generate_output_file_hashes,
     )
-    from mtgjson5.price_builder import build_prices
     from mtgjson5.providers import GitHubMTGSqliteProvider, ScryfallProvider
 
     # If a price build, simply build prices and exit
     if args.price_build:
-        generate_compiled_prices_output(build_prices(), args.pretty)
+        build_price_files(args.pretty)
         if args.compress:
             compress_mtgjson_contents(MtgjsonConfig().output_path)
         generate_output_file_hashes(MtgjsonConfig().output_path)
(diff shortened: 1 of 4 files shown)


kodawah:
couple of comments, looking at the output file provided

1. 
>                         "foil": {},

I think it'd be better to drop empty prices structs

2. how are etched cards marked?
3. we should normalize the finish names with upstream `nonfoil` and `foil` IMO
4. the `currency` field does not belong to the price structure - if it's not for `meta` it should have a separate body at the same level of `data`
5. make sure memory does not grow too much during execution
6. what's the point of keeping the date?

ZeldaZach:
The original idea was to create a "smaller" version of the already existing file. While I agree there are some design choices that can be improved upon, I'd prefer to hold them off until we do a complete overhaul of the prices struct (Maybe we build sub-versions to keep compatibility?)

This would allow people to drag/drop in the AllPricesToday file as a replacement and get the data they need @kodabb

kodawah:
> This would allow people to drag/drop in the AllPricesToday file as a replacement and get the data they need

i don't think creating a future technical debt is a good idea

ZeldaZach:
We decided against this idea, and found an alternative. Still want to plan the ideas for a smaller file to distribute.

PR #971, opened 2022-10-19
Align and order TcgplayerSkus enums to proper name

staghouse:
Fixes EnumValues being set for TcgplayerSkus using incorrect property names.


ZeldaZach (approved):
Thanks!

PR #987, opened 2022-12-09
Fix #978: Link related cards in the same set

### LINKED ISSUE #978: implement related foil field
It would be nice to have a way to map cards that have foil cards listed in separate entries: this is especially useful for 10E and UNH cards that have a single `finish` entry but cards are actually coupled by their collector number.

**proposal**
Introduce two fields that keep track of this mapping in the Card object
- identifiers.FoilOf: uuid
- identifiers.NonfoilOf: uuid

**implementation** *pseudocode*

for set in "Battlebond",	"Conspiracy: Take the Crown", "Fate Reforged", "Onslaught", "Tenth Edition", "Unhinged":
     if card.illustrationid == other_card.illustrationid then
         if card.isFoil()
              card.identifiers.FoilOf = other_card.uuid
         else if card.isNonfoil()
              card.identifiers.NonfoilOf = other_card.uuid

this loop is simple enough that it could be performed for any set, however there might be a few outliers that need manual mapping too

ZeldaZach:
- ADD Identifiers.mtgjsonNonFoilVersionId
- ADD Identifiers.mtgjsonFoilVersionId

Not fully certain if I want this or not, but it was simple to POC so I did that
cc @kodabb 

[10E.json.txt](https://github.com/mtgjson/mtgjson/files/10191098/10E.json.txt)

### DIFF
--- a/mtgjson5/set_builder.py
+++ b/mtgjson5/set_builder.py
@@ -431,6 +431,9 @@ def build_mtgjson_set(set_code: str) -> Optional[MtgjsonSetObject]:
     add_mcm_details(mtgjson_set)
     add_card_kingdom_details(mtgjson_set)
 
+    if mtgjson_set.code in {"CN2", "FRF", "ONS", "10E", "UNH"}:
+        link_same_card_different_details(mtgjson_set)
+
     if mtgjson_set.code in {"EMN", "BRO"}:
         add_meld_face_parts(mtgjson_set)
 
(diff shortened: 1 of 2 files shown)


mtgjson5/set_builder.py
```diff
@@ -1185,6 +1188,36 @@ def add_other_face_ids(cards_to_act_on: List[MtgjsonCardObject]) -> None:
...
+
+    for mtgjson_card in mtgjson_set.cards:
+        if mtgjson_card.identifiers.scryfall_illustration_id not in cards_seen:
+            cards_seen[mtgjson_card.identifiers.scryfall_illustration_id] = mtgjson_card
+            continue
+
+        other_mtgjson_card = cards_seen[
+            mtgjson_card.identifiers.scryfall_illustration_id
+        ]
+        if "nonfoil" in mtgjson_card.finishes:
```
kodawah, line 1209: should you check that nonfoil is the only one too?

ZeldaZach: For the time being, it's not necessary as all of the cases are a foil vs non-foil discrepancy, which lead to the splitting

PR #995, opened 2022-12-21
Fix #643: ADD MtgjsonSet.Languages List of Strings

### LINKED ISSUE #643: Available languages for printed sets

ZeldaZach:
[ORI.json.txt](https://github.com/mtgjson/mtgjson/files/10279970/ORI.json.txt)

### DIFF
--- a/mtgjson5/providers/scryfall/monolith.py
+++ b/mtgjson5/providers/scryfall/monolith.py
@@ -11,10 +11,10 @@ from typing import Any, Dict, List, Optional, Set, Union
 import ratelimit
 from singleton_decorator import singleton
 
-from .. import constants
-from ..mtgjson_config import MtgjsonConfig
-from ..providers.abstract import AbstractProvider
-from ..utils import retryable_session
+from ... import constants
+from ...mtgjson_config import MtgjsonConfig
+from ...providers.abstract import AbstractProvider
+from ...utils import retryable_session
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


mtgjson5/providers/scryfall/monolith.py
```diff
@@ -11,10 +11,10 @@
...
 from singleton_decorator import singleton
 
-from .. import constants
-from ..mtgjson_config import MtgjsonConfig
-from ..providers.abstract import AbstractProvider
-from ..utils import retryable_session
+from ... import constants
+from ...mtgjson_config import MtgjsonConfig
+from ...providers.abstract import AbstractProvider
+from ...utils import retryable_session
```
kodawah, line 17: 💠

kodawah (approved):
seems good, only improvement/change i'd make is exposing the standard language tag instead of the English one, but that can be separate i suppose

mtgjson5/providers/scryfall/monolith.py
ZeldaZach: I don't know what that emote means

ZeldaZach:
Everything in MTGJSON has been through the English Tag compared to standard one, as it's not standard across providers

mtgjson5/providers/scryfall/monolith.py
kodawah: it's a dot, a diamond dot

Issue #799, opened 2021-05-13
future sealed work


kodawah:
tracking bug for some sealed work I'd like to see before public release

- [x] if appropriate, add number of boosters in the box
- [x] add a category field - we need to be able to understand whether a product is a booster box/booster pack/deck/special set
- [x] add a decklist reference - for set with defined decklists we need references to the cards contained or links to additional resources
- [ ] add an image link - so that sites can publish the data
- [x] add missing product - currently tcg covers most products, but there are a bunch of missing ones
  - [ ] a bunch of old theme decks (visions, morningtide etc)
  - [ ] various starter decks and intro packs
  - [ ] a lot of sealed boxes, such as mythic edition, theme deck boxes
- [ ] add missing editions - some editions do not map easily to scryfall models, in particular 
  - [ ] https://www.tcgplayer.com/search/all/product?q=Duels%20of%20the%20Planeswalkers 
  - [ ] https://www.tcgplayer.com/search/magic/world-championship-decks?productLineName=magic&setName=world-championship-decks&ProductTypeName=Sealed%20Products&page=1
  - [ ] https://www.tcgplayer.com/search/magic/spellslinger-starter-kit?productLineName=magic&setName=spellslinger-starter-kit
  - [ ] https://www.tcgplayer.com/search/magic/challenger-decks?productLineName=magic&setName=challenger-decks
  - [ ] https://www.tcgplayer.com/search/magic/gift-boxes-and-promos?productLineName=magic&setName=gift-boxes-and-promos
  - [ ] https://shop.tcgplayer.com/magic/mythic-edition-war-of-the-spark/war-of-the-spark-mythic-edition
  - [ ] https://shop.tcgplayer.com/magic/mythic-edition-ravnica-allegiance/ravnica-allegiance-mythic-edition
- [x] fixup a few typos and improve naming consistency
  - [x] "Devestation", "Rambler", "Artufl"
  - [x] a bunch of decks use the color instead of the actual deck name
  - [x] intro pack sets sometimes use "all 5 intro packs" and sometimes use "set of 5"
  - [x] expose `productSize` for `booster` product types too

staghouse:
assuming you're talking about `sealedProduct` - `releaseDate` may return null, is this intentional?

kodawah:
@staghouse yes i'm referring to the new element. I think `releaseDate` is just what tcg reports, we probably should strip empty values though (cc @tpblaster)

taw:
Is this about decklists?

[...]

PR #997, opened 2022-12-22
Set boosterBoxSize for relevant sealedProduct entries

kodawah:
part of #799


mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +276,33 @@ def generate_mtgjson_sealed_product_objects(
...
+            # XXX: this should be wrapped up in a category field
+            if "booster box" in product["cleanName"].lower():
+                if "set" in product["cleanName"].lower():
+                    sealed_product.booster_box_size = 30
+                    if code in booster_box_size_overrides["set"].keys():
+                        sealed_product.booster_box_size = int(booster_box_size_overrides["set"][code])
+                else if "collector" in product["cleanName"].lower():
+                    sealed_product.booster_box_size = 12
+                    if code in booster_box_size_overrides["collector"].keys():
+                        sealed_product.booster_box_size = int(booster_box_size_overrides["collector"][code])
```
ZeldaZach: `sealed_product.booster_box_size = int(booster_box_size_overrides.get("collector",{}).get(code, 12)`

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +276,33 @@ def generate_mtgjson_sealed_product_objects(
 
         mtgjson_sealed_products = []
 
+        # Load cache if not loaded
+        with constants.RESOURCE_PATH.joinpath("booster_box_size_overrides.json").open(encoding="utf-8") as f:
+            booster_box_size_overrides = json.load(f)
+
+        code = self.code.upper()
```
ZeldaZach: There is no `self.code` in this class, you'll need to pass `set_code: str` in to the function

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +276,33 @@ def generate_mtgjson_sealed_product_objects(
...
+
         for product in sealed_data:
             sealed_product = MtgjsonSealedProductObject()
             sealed_product.name = product["cleanName"]
             sealed_product.identifiers.tcgplayer_product_id = str(product["productId"])
             sealed_product.release_date = product["presaleInfo"].get("releasedOn")
+
+            # XXX: this should be wrapped up in a category field
+            if "booster box" in product["cleanName"].lower():
```
ZeldaZach: This block should be broken out into its own function

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +276,33 @@ def generate_mtgjson_sealed_product_objects(
...
+                if "set" in product["cleanName"].lower():
+                    sealed_product.booster_box_size = 30
+                    if code in booster_box_size_overrides["set"].keys():
```
ZeldaZach: `.keys()` isn't necessary anymore, as it's default behavior (Mypy should've caught this)

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +276,33 @@ def generate_mtgjson_sealed_product_objects(
...
+
         for product in sealed_data:
             sealed_product = MtgjsonSealedProductObject()
             sealed_product.name = product["cleanName"]
             sealed_product.identifiers.tcgplayer_product_id = str(product["productId"])
             sealed_product.release_date = product["presaleInfo"].get("releasedOn")
+
+            # XXX: this should be wrapped up in a category field
+            if "booster box" in product["cleanName"].lower():
+                if "set" in product["cleanName"].lower():
```
ZeldaZach: `if "Set" in product["cleanName'}` removes the need to lower (same elsewhere)

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +276,33 @@ def generate_mtgjson_sealed_product_objects(
 
         mtgjson_sealed_products = []
 
+        # Load cache if not loaded
+        with constants.RESOURCE_PATH.joinpath("booster_box_size_overrides.json").open(encoding="utf-8") as f:
```
ZeldaZach: File not found :O

kodawah:
[ATQ.json.txt](https://github.com/mtgjson/mtgjson/files/10285547/ATQ.json.txt)
[IKO.json.txt](https://github.com/mtgjson/mtgjson/files/10285549/IKO.json.txt)
[ZNR.json.txt](https://github.com/mtgjson/mtgjson/files/10285550/ZNR.json.txt)

👍

mtgjson5/providers/tcgplayer.py
```diff
@@ -256,8 +269,178 @@ def generate_today_price_dict(
...
+        self, product_name: str
+    ) -> MtgjsonSealedProductCategory:
+        """
+        Best-effort to parse the product name and determine the sealed product category
+        :param product_name Name of the product from TCG
+        :return: category
+        """
+
+        # The order of the checks is important
+        for tag in [
```
ZeldaZach: Could simplify this with `if any(tag for tag in list): return CASE`

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +459,53 @@ def generate_mtgjson_sealed_product_objects(
...
+            encoding="utf-8"
+        ) as f:
+            booster_box_size_overrides = json.load(f)
+
+        with constants.RESOURCE_PATH.joinpath("sealed_name_fixup.json").open(
+            encoding="utf-8"
+        ) as f:
+            sealed_name_fixup = json.load(f)
+
+        with constants.RESOURCE_PATH.joinpath("booster_box_size_overrides.json").open(
```
ZeldaZach: this is duplicated of 462

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +459,53 @@ def generate_mtgjson_sealed_product_objects(
 
         mtgjson_sealed_products = []
 
+        with constants.RESOURCE_PATH.joinpath("booster_box_size_overrides.json").open(
+            encoding="utf-8"
+        ) as f:
+            booster_box_size_overrides = json.load(f)
```
ZeldaZach: might not be a good idea to call it booster box, since this is every type of product now?

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +459,53 @@ def generate_mtgjson_sealed_product_objects(
...
+            )
+            sealed_product.subtype = self.determine_mtgjson_sealed_product_subtype(
+                sealed_product.name.lower(), sealed_product.category
+            )
+
+            LOGGER.debug(
+                f"{sealed_product.name}: {sealed_product.category.value}.{sealed_product.subtype.value}"
+            )
+
+            if sealed_product.category == MtgjsonSealedProductCategory.BOOSTER_BOX:
```
ZeldaZach, line 513: Is this override really needed here?

mtgjson5/classes/mtgjson_sealed_product.py
```diff
@@ -1,13 +1,117 @@
...
 """
+import enum
 from typing import Any, Dict, Optional
 
 from ..utils import to_camel_case
 from .mtgjson_identifiers import MtgjsonIdentifiersObject
 from .mtgjson_purchase_urls import MtgjsonPurchaseUrlsObject
 
 
+class MtgjsonSealedProductCategory(enum.Enum):
```
ZeldaZach, line 12: I'd prefer each of these enums to be defined in their own file, and then use the `__init__` to bring them together

mtgjson5/providers/tcgplayer.py
```diff
@@ -100,6 +106,13 @@ class TCGPlayerProvider(AbstractProvider):
         "3x Magic Booster Packs",
         "Booster Battle Pack",
     ]
+    booster_box_default_size = {
```
ZeldaZach: not sure if booster_box is the correct term anymore

mtgjson5/providers/tcgplayer.py
kodawah: ok

mtgjson5/providers/tcgplayer.py
kodawah: fixed

mtgjson5/providers/tcgplayer.py
kodawah: yes sometimes "draft" is not present so we need to fallback on what we know

kodawah: i'll add a comment

kodawah: oh that's not an override, it's just that the data we have only covers the booster box size

mtgjson5/providers/tcgplayer.py
kodawah: ok!

mtgjson5/providers/tcgplayer.py
kodawah: it's not, moving to product_default_size, but for now not converting it to a dict

mtgjson5/classes/mtgjson_sealed_product.py
kodawah: i think it makes sense to keep everything related to sealed in a single place no?

mtgjson5/providers/tcgplayer.py
```diff
@@ -256,8 +269,194 @@ def generate_today_price_dict(
...
+        """
+        Best-effort to parse the product name and determine the sealed product category
+        :param product_name Name of the product from TCG
+        :return: category
+        """
+
+        # The order of the checks is important
+        if any(
+            tag in product_name
+            for tag in [
```
ZeldaZach, line 287: Would it make sense to const-ify these lists across the board? Would simplify the code here and maybe let us lambda function this

mtgjson5/providers/tcgplayer.py
```diff
@@ -256,8 +269,194 @@ def generate_today_price_dict(
...
+    def determine_mtgjson_sealed_product_category(
+        self, product_name: str
+    ) -> MtgjsonSealedProductCategory:
+        """
+        Best-effort to parse the product name and determine the sealed product category
+        :param product_name Name of the product from TCG
+        :return: category
+        """
+
+        # The order of the checks is important
```
ZeldaZach: I feel this comment is underrated and needs more context to it on why the order matters

mtgjson5/providers/tcgplayer.py
kodawah: i don't think so? these are just random names that happened to be on tcg and could change at any moment

what did you have in mind?

mtgjson5/providers/tcgplayer.py
kodawah: ill expand on it

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +475,48 @@ def generate_mtgjson_sealed_product_objects(
...
+                f"{sealed_product.name}: {sealed_product.category.value}.{sealed_product.subtype.value}"
+            )
+
+            if sealed_product.category == MtgjsonSealedProductCategory.BOOSTER_BOX:
+                sealed_product.product_size = int(
+                    booster_box_size_overrides.get(
+                        sealed_product.subtype.value, {}
+                    ).get(
+                        set_code,
+                        self.product_default_size.get(sealed_product.subtype.value, {}),
```
ZeldaZach: Getting a default dict if the sealed_product isn't found doesn't seem right. Should be an integer

mtgjson5/resources/sealed_name_fixes.json
```diff
@@ -0,0 +1,12 @@
+{
+    "DotP": "Duels of the Planeswalkers",
+    "M15 Prerelease": "Magic 2015 Prerelease",
+    "All 5": "Set of 5",
+    "M10 ": "",
```
ZeldaZach, line 5: Is it intentional the M1X have a space after the name

mtgjson5/providers/tcgplayer.py
ZeldaZach: Just a general idea I was thinking

```python3
def helper_with_better_name(find_list, return_value_if_found):
    return value_if_found if any(x for x in find_list) else None
```

and just repeatedly call to that setup. Could also be a lambda
```python3
lambda x, y: return y if any(a for a in x) else None
```

kodawah: i might need an example for this ^^

mtgjson5/resources/sealed_name_fixes.json
kodawah: yes, it's needed to remove an extra space

mtgjson5/providers/tcgplayer.py
kodawah: good catch

mtgjson5/resources/sealed_name_fixes.json
kodawah: if you want i can just remove the space in the tag and then add an extra entry to replace `"  "` with `" "`

tpblaster:
I am a fan of these changes, with these combined with work from Taw, we should be close to having a solid base for sealed products.

mtgjson5/providers/tcgplayer.py
```diff
@@ -276,11 +478,48 @@ def generate_mtgjson_sealed_product_objects(
 
         mtgjson_sealed_products = []
 
+        with constants.RESOURCE_PATH.joinpath("sealed_name_fixes.json").open(
+            encoding="utf-8"
+        ) as f:
+            sealed_name_fixes = json.load(f)
+
+        with constants.RESOURCE_PATH.joinpath("sealed_name_fixes.json").open(
```
ZeldaZach: This is loading the wrong file
```suggestion
        with constants.RESOURCE_PATH.joinpath("booser_box_size_overrides.json").open(
```

ZeldaZach (approved):

PR #1024, opened 2023-01-26
Minor Bug Fixes and Added Testing for TCGPlayer Sealed Product Type Detection

tpblaster:
This PR adds minor bug fixes and testing for the function that detects the type of a sealed product based on its name.

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -7,6 +7,7 @@ pytest-cov~=4.0.0
 pytest-mock~=3.10.0
 setuptools~=66.0.0
 tox~=3.28.0
+pandas~=1.5.3
 
 types-PyYAML~=6.0.12.3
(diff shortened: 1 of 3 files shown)


kodawah (approved):
i approve of more pandas being deployed

requirements_test.txt
```diff
@@ -7,6 +7,7 @@ pytest-cov~=4.0.0
 pytest-mock~=3.10.0
 setuptools~=66.0.0
 tox~=3.28.0
+pandas~=1.3.5
```
ZeldaZach: pandas already exists in our [requirements.txt](https://github.com/mtgjson/mtgjson/blob/master/requirements.txt#L9) file. This shouldn't be necessary

kodawah: zach hates pandas confirmed

PR #1061, opened 2023-04-22
Sealed products override

axxroytovu:
Sets up a template and structure for sealed product overrides.

Manual products and product contents are hosted [here](https://github.com/axxroytovu/mtg-sealed-contents).

Additional sealed product information is downloaded from Card Kingdom.


axxroytovu:
Putting this up now if anyone has comments or improvements to the workflow before I finalize the pull.

axxroytovu:
Reworked everything to use a GitHub provider similar to how boosters work. Just need to test it later. 

Edit: Well, I don't have any personal API tokens so I'm going to submit it and people can tell me how it breaks everything.

kodawah:
i think it should be squashed to a single commit, but overall looks good
will the additional_sealed_product file be added as a later stage?
if you're adding ck purchase urls, maybe we could add an identifiers map too?

axxroytovu:
> will the additional_sealed_product file be added as a later stage? 

I'm hosting the additional sealed products here: https://github.com/axxroytovu/mtg-sealed-contents/blob/main/outputs/products.json

> if you're adding ck purchase urls, maybe we could add an identifiers map too?

That sounds like a great idea, I'll add that.

axxroytovu:
Ok, never mind. I have no idea how to find the CK identifiers for sealed product. It's nowhere on their website and I can't find anything in their API that would expose them. 

I've at least allowed the code to read location IDs provided by the override json.

tpblaster:
The CK API for sealed products is: https://api.cardkingdom.com/api/sealed_pricelist

axxroytovu:
Thank you!

axxroytovu:
Ok, I think loading from Card Kingdom is working. I can't test any further without a TCGPlayer API key, but it at least won't crash with the current setup. There will be a lot of duplicate products until I can make a map of TCGPlayer product names to Card Kingdom product names.

axxroytovu:
@kodabb Overrides from remote host can now include a generic "product size" to resolve newest issue in https://github.com/mtgjson/mtgjson/issues/799

kodawah:
> @kodabb Overrides from remote host can now include a generic "product size" to resolve newest issue in https://github.com/mtgjson/mtgjson/issues/799

so there is something i haven't thought of, what's the unit of measure for this field?
Are we supposed to keep a different map for each product type (15 cards for booster, 36 for booster boxes etc) or introduce a common unit size in terms of "booster count"?

axxroytovu:
Each individual product has its own "product size" defined. Booster boxes can each have different sizes. Sizes correspond to the main "contents" of the box. So a booster box for example:
```yaml
  The Brothers War Jumpstart Booster Display:
    sealed:
    - set: bro
      name: The Brothers War Jumpstart Booster Pack
      count: 18
    size: 18
```
or a case:
```yaml
  The Brothers War Set Booster Display Case:
    sealed:
    - set: bro
      name: The Brothers War Set Booster Display
      count: 6
    size: 6
```
or a booster pack:
```yaml
  The Brothers War Collector Booster Pack:
    pack:
    - set: bro
      code: collector
    size: 15
```

kodawah:
yeah but on mtgjson side `productSize` will mean three different things
- the number of boosters in booster_box and draft_set
- the number of cards in booster_pack and deck
- the number of boxes in case
- something else for other categories

it seems a bit of overloaded API to be honest, maybe having different fields for different things would be better (or at least drop productSize and use a map for each field) - what do you think?

axxroytovu:
That's absolutely fair. I personally am in favor of deprecating `product_size` entirely in favor of the new `contents` system but people wanted to keep it around.

If we deprecated `product_size`, then we could just add something like `card_count` to the base level products: packs and decks. That would look like:

```yaml
Brothers War Draft Booster Box:
  sealed:
  - name: Brothers War Draft Booster
    set: bro
    count: 36
Brothers War Draft Booster:
  pack:
  - set: bro
    code: default
  card_count: 15
```

Does that solve the problem?

ZeldaZach:
I'm doing some looking on this, and I appreciate your insights! @axxroytovu I'm open with deprecating product_size if necessary to have a more clear system.

In addition, I'm a bit concerned with having an extra 3rd party dependency, and would rather host this kind of override information within a mtgjson organization repository, if possible. 

If you need help testing your product, I have keys for all of these services and can provide you with responses or dumps from your code. Feel free to chat with me more on Discord.

axxroytovu:
I'm absolutely happy to have MTGJson host the repo. It's two python files that just make data entry more user friendly and a bunch of YAML files. If you want to fork/copy over what I have we can make the MTGJson source the main one.

mallardduck:
> yeah but on mtgjson side `productSize` will mean three different things
> 
> * the number of boosters in booster_box and draft_set
> * the number of cards in booster_pack and deck
> * the number of boxes in case
> * something else for other categories

This was kinda the assumption I already had going into my other comments and requests. Within my own app I was already considering that I had to determine a products `category` (and sometimes `subtype`) to know what base "unit" the `productSize` was describing.

However that is a lot more effort than the `contents` mechanism described by @axxroytovu and the example they provided. So I would totally be in favor of moving in that direction too! It seems this would provide me with the counts I'm hoping for, so no complaint about that.

axxroytovu:
@ZeldaZach Everything is updated and working with the new remote location.

The one improvement I still want to make but don't know how is to simplify the "promo card" callouts. Right now they resolve like this (example is the Brothers' War Bundle):
```json
      "contents": {
          "card": [
              {
                  "foil": true,
                  "name": "Queen Kayla bin-Kroog",
                  "number": 379,
                  "set": "bro"
              }
          ]}
```
This could realistically be reduced to just the UUID without any loss of information, but I don't know how to make that swap. This isn't a big deal and could probably be an improvement later but I wanted to bring it up.

mtgjson5/resources/sealed_name_fixes.json
```diff
@@ -1,15 +1,55 @@
...
+    "Magic 2012": "2012 Core Set",
+    "Magic 2013": "2013 Core Set",
+    "Magic 2014": "2014 Core Set",
+    "Magic 2015": "2015 Core Set",
+    "The Brothers War": "The Brothers' War",
+    "Shards of Alara Block": "Shards of Alara",
+    "Set of Two": "Set of 2",
+    "[Set of 2]": "Set of 2",
+    "Decks - Set": "\u2691",
+    "Decks Set": "\u2691",
```
kodawah, line 43: this is a bit confusing IMO, I'd just repeat "Decks - Set" for each

mtgjson5/resources/cardkingdom_sealed_name_mapping.json
```diff
@@ -0,0 +1,204 @@
+{
+    "editions": {
+        "2017 gift pack": "gift pack 2017",
```
kodawah, line 3: could we make this json as a single children type? right now you can have string:string or string:array_of_strings which is confusing and hard to unmarshal - making it always string:array_of_string makes it simpler IMO

mtgjson5/resources/sealed_name_fixes.json
axxroytovu: What this block is actually cleaning up is "deck" vs "decks". Card Kingdom likes to use `{set} commander decks {name}` where TCGplayer uses `{set} commander deck {name}`.

At the same time, I can't just replace "decks" with "deck" globally because then `commander decks set of 2` type products look like a mess. 

This bit of code finds anywhere that is a set of decks, turns that phrase into a searchable flag character, then deals with anywhere that Deck is mutable, then turns the flag character back into `decks - set`.

kodawah: hmm yeah it's a mess xD
could it at least be documented somewhere?

axxroytovu: That's the plan eventually. I wish JSON allowed for comments.

kodawah:
I found one more name that needs a change

sealedProduct:
> Adventures in the Forgotten Realms Theme Booster Dungeon Theme

booster:
> Adventures in the Forgotten Realms Theme Booster Dungeons

mtgjson5/resources/sealed_name_fixes.json
kodawah: what i do for json comments is makeup my own entry on top of the necessary one, so like

```json
    "[Set of 2]": "Set of 2",
    "COMMENT1": "this is because xxxx",
    "Decks - Set": "\u2691",
    "Decks Set": "\u2691",
```

not pretty but gets the job done

axxroytovu: Comments added in latest commit

mtgjson5/resources/cardkingdom_sealed_name_mapping.json
axxroytovu: I ended up removing the arrays because it was a hokey fix for something that should be resolved on a bigger level. See https://github.com/mtgjson/mtgjson/issues/1070.

axxroytovu:
Ok, the majority of work here should be done. Duplicates from tcgplayer/cardkingdom have been mapped and should parse correctly. Still missing some products, see https://github.com/mtgjson/mtgjson/issues/1070 .

axxroytovu:
> I found one more name that needs a change

I believe the product name is correct and this should get changed on the booster side if anything. I don't think this is too big of a deal but should be a fairly easy fix if needed.

ZeldaZach:
@axxroytovu Will need to fix the failing tests before I can merge

axxroytovu:
> @axxroytovu Will need to fix the failing tests before I can merge

That should fix it

axxroytovu:
This error keeps showing up in the tests and I don't know how to resolve it: 
`mtgjson5/providers/cardkingdom.py:353: error: Cannot determine type of "sealed_product"  [has-type]`
Any help is appreciated.

mallardduck:
> This error keeps showing up in the tests and I don't know how to resolve it: `mtgjson5/providers/cardkingdom.py:353: error: Cannot determine type of "sealed_product" [has-type]` Any help is appreciated.

I think it's because how you are using the variable `sealed_product`. It's first referenced on line 348 before it has a defined value. And is set to a value at either 350, or 355 depending on things - and those don't have consistent types. Setting the variable at 350 is an ambiguous type, but at 355 it's implicitly a `MtgjsonSealedProductObject` type.

So I think to make that test happy you may just have to type hint the variable?

> sealed_product: MtgjsonSealedProductObject = existing_names[check_name]

Or maybe rather it needs the hint as a comment:

> sealed_product = existing_names[check_name]  # type: MtgjsonSealedProductObject

TBH, I'm a bit rusty on how python 3 does the types stuff.

mtgjson5/resources/cardkingdom_sealed_name_mapping.json
```diff
@@ -0,0 +1,307 @@
...
+        "vanguard series": "vanguard",
+        "unlimited edition": "unlimited",
+        "commander 2011": "commander",
+        "commander anthology volume ii": "commander anthology vol. ii",
+        "zendikar rising commander": "zendikar rising commander decks",
+        "eighth edition": "8th edition",
+        "conspiracy: take the crown": "conspiracy - take the crown",
+        "limited edition alpha": "alpha",
+        "magic 2011": "2011 core set",
+        "kaldheim": ["kaldheim", "challenger decks 2021"],
```
kodawah, line 47: there is a leftover array here

mtgjson5/resources/cardkingdom_sealed_name_mapping.json
```diff
@@ -0,0 +1,307 @@
...
+        "magic origins intro pack hunting pack blackgreen": "magic origins intro pack hunting pack",
+        "zendikar intro pack kor armory": "zendikar intro pack white kor armory",
+        "zendikar intro pack pumped up": "zendikar intro pack red pumped up",
+        "zendikar intro pack rise of the vampires": "zendikar intro pack black rise of the vampires",
+        "zendikar intro pack unstable terrain": "zendikar intro pack blue unstable terrain",
+        "zendikar intro pack the adventurers": "zendikar intro pack green the adventurers",
+        "odyssey theme deck trounceomatic": "odyssey theme deck trounce o matic",
+        "odyssey theme deck onetwo punch": "odyssey theme deck one two punch",
+        "warhammer 40000 commander deck forces of the imperium": "universes beyond warhammer 40000 forces of the imperium commander deck",
+        "warhammer 40000 commander deck tyranid swarm": "universes beyond warhammer 40000 tyranid swarm commander deck",
```
kodawah, line 299: for the future, we could probably strip "universes beyond" tags, since they are likely to be reused

PR #1065, opened 2023-04-25
Override ONC base set size

### LINKED ISSUE #1060: ONE baseSetSize seems incorrect
5.2.1+20230420

#### Description of Bug

It calculated 284, when in reality it should be 271.
Maybe the promotypes "promopack" and/or "bundle" should be included when checking "boosterfun" in the `set_builder.py`.

The 284 comes from including the following blocks of cards.

272-276: Basic lands not counted towards the main set https://scryfall.com/card/one/272/plains
They have the promotype "bundle"

277-283: Promo packs: https://scryfall.com/card/one/277/ossification

davidmartos96:
Fixes https://github.com/mtgjson/mtgjson/issues/1060#issuecomment-1519590719


mallardduck (approved):
LGTM

PR #1066, opened 2023-04-27
Add MOM Aftermath epilogue subtype and sizes

mallardduck:
WOTC claims that the Epilogue is a new product type (doubt) so we might expect more of them in the future. With this current set there is no "draft" (aka default) product, nor set. So I've added the new subtype and an override for the collector box which also has less cards than regular Set collector boxes.

I do not have any valid API keys to the data sources - as such I cannot test this change fully since the sealedProduct data comes from TCGplayer and I cannot fetch that data.


kodawah:
the booster is still a "default" one for that set and it matches the name of the sheet, so even if it's a new product type, it doesn't need to have that booster type in sealed data. Instead I'd consider a new set.Type value, but that depends on upstream (scryfall)

mtgjson5/resources/booster_box_size_overrides.json
```diff
@@ -34,7 +34,8 @@
         "CLB": 18
     },
     "collector": {
-        "2X2": 6
+        "2X2": 6,
+        "MAT": 12
```
mallardduck, line 38: @kodabb - i closed this as I'm obviously lacking context about how specific things are used. However I would suspect this item is still needed as it seems similar to the 2X2 situation.

kodawah: unless I'm mistaken 12 is the default value for collector boosters, but the "default" one does need to be set to 24, will you send a pr for it?

mallardduck: Oh yeah that's correct - the number of packs by default for collector boxes is 12.

However the number of cards per pack is what's odd about MAT, I must have been thinking that.
Normal collector boosters have 15 cards, MAT only have 6 cards per pack.

PR #1071, opened 2023-05-09
Prevent aliasing sealed subtype with Core Set editions

kodawah:
otherwise all M21/M20 etc products get the 'set' subtype


ZeldaZach:
In the future, I'd like to see unit tests supporting this kind of stuff please

PR #1083, opened 2023-06-07
Support Decks inside of SET.json/AllPrintings

### LINKED ISSUE #1076: Support Decks inside of Set Files & AllPrintings
It would be nice to have decklists embedded in the main AllPrintings data model for ease of access when determining ev and other properties. I see a couple of ways to do this

1. add the current decklists in the set data
2. add a fixed sheet in Booster, with custom Contents
3. add a separate data format for decks with just a (fixed) sheet

I'm favoring #2 because it would allow to reuse parsing format used for normal products, but I'd like to hear more thoughts - cc @ZeldaZach @taw @axxroytovu

ZeldaZach:
- Fix #1076
- UUIDs link back to sealedProduct, if able

### DIFF
--- a/mtgjson5/classes/__init__.py
+++ b/mtgjson5/classes/__init__.py
@@ -21,4 +21,5 @@ from .mtgjson_sealed_product import (
     MtgjsonSealedProductSubtype,
 )
 from .mtgjson_set import MtgjsonSetObject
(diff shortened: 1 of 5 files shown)


ZeldaZach:
This still has room for improvement, as non-decks are in the decks array. Might need to filter on them

PR #1098, opened 2023-07-04
Empty sets

axxroytovu:
Add code to build specific sets that do include decks, boosters, and sealed product but no cards

### DIFF
--- a/mtgjson5/__main__.py
+++ b/mtgjson5/__main__.py
@@ -102,6 +102,13 @@ def dispatcher(args: argparse.Namespace) -> None:
         return
 
     sets_to_build = ScryfallProvider().get_sets_to_build(args)
+    if args.all_sets:
+		with constants.RESOURCE_PATH.joinpath("additional_sets.json").open(
+			encoding="utf-8"
+		) as f:
+			additional_set_keys = set(json.load(f).keys())
+			additional_set_keys -= set(args.skip_sets)
+			sets_to_build += list(additional_set_keys)
     if sets_to_build:
         build_mtgjson_sets(sets_to_build, args.pretty, args.referrals)
 
--- a/mtgjson5/__main__.py
+++ b/mtgjson5/__main__.py
@@ -109,7 +109,7 @@ def dispatcher(args: argparse.Namespace) -> None:
         ) as f:
             additional_set_keys = set(json.load(f).keys())
             additional_set_keys -= set(args.skip_sets)
-            sets_to_build += list(additional_set_keys)
+            sets_to_build = list(set(sets_to_build) + additional_set_keys)
     if sets_to_build:
         build_mtgjson_sets(sets_to_build, args.pretty, args.referrals)
 
--- a/mtgjson5/__main__.py
+++ b/mtgjson5/__main__.py
@@ -109,7 +109,7 @@ def dispatcher(args: argparse.Namespace) -> None:
         ) as f:
             additional_set_keys = set(json.load(f).keys())
             additional_set_keys -= set(args.skip_sets)
-            sets_to_build = list(set(sets_to_build) + additional_set_keys)
+            sets_to_build = list(set(sets_to_build).union(additional_set_keys))
     if sets_to_build:
         build_mtgjson_sets(sets_to_build, args.pretty, args.referrals)
 
--- a/mtgjson5/__main__.py
+++ b/mtgjson5/__main__.py
@@ -6,7 +6,6 @@ import gevent.monkey  # isort:skip
 gevent.monkey.patch_all()  # isort:skip
 
 import argparse
-import json
 import logging
 import traceback
 from typing import List, Set, Union
@@ -14,7 +13,7 @@ from typing import List, Set, Union
 import urllib3.exceptions
 
 from mtgjson5 import constants
-from mtgjson5.utils import init_logger
+from mtgjson5.utils import init_logger, load_local_set_data
 
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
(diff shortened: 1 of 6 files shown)


mtgjson5/set_builder.py
```diff
@@ -404,7 +404,13 @@ def build_mtgjson_set(set_code: str) -> Optional[MtgjsonSetObject]:
     # Ensure we have a header for this set
     set_data = get_scryfall_set_data(set_code)
     if not set_data:
-        return None
+        with constants.RESOURCE_PATH.joinpath("additional_sets.json").open(
+            encoding="utf-8"
+        ) as f:
+            additional_sets_data = json.load(f)
+            set_data = additional_sets_data.get(set_code.lower(), False)
```
ZeldaZach: No need to add a False, as the default of `get(key)` is None. Also, I like using set_code UPPER if possible across the code base

mtgjson5/__main__.py
```diff
@@ -102,6 +103,13 @@ def dispatcher(args: argparse.Namespace) -> None:
         return
 
     sets_to_build = ScryfallProvider().get_sets_to_build(args)
+    if args.all_sets:
+        with constants.RESOURCE_PATH.joinpath("additional_sets.json").open(
```
ZeldaZach: This code is used in two places, can we abstract it out?

mtgjson5/__main__.py
```diff
@@ -102,6 +103,13 @@ def dispatcher(args: argparse.Namespace) -> None:
...
+            encoding="utf-8"
+        ) as f:
+            additional_set_keys = set(json.load(f).keys())
+            additional_set_keys -= set(args.skip_sets)
+            sets_to_build += list(additional_set_keys)
```
ZeldaZach: I'm not sure if loading, subtracting, and adding is correct as there could be duplicates.

mtgjson5/providers/cardkingdom.py
```diff
@@ -178,11 +175,15 @@ def update_sealed_product(
 
             sealed_product.identifiers.card_kingdom_id = str(product["id"])
 
-            sealed_product.category = sealed_product.determine_mtgjson_sealed_product_category(
-                sealed_product.name.lower()
+            sealed_product.category = (
```
ZeldaZach: Did tox clean this file up here? Feels unmodified

mtgjson5/providers/tcgplayer.py
```diff
@@ -333,11 +333,15 @@ def generate_mtgjson_sealed_product_objects(
             sealed_product.identifiers.tcgplayer_product_id = str(product["productId"])
             sealed_product.release_date = product["presaleInfo"].get("releasedOn")
 
-            sealed_product.category = sealed_product.determine_mtgjson_sealed_product_category(
-                sealed_product.name.lower()
+            sealed_product.category = (
```
ZeldaZach: This file should probably be left as it was, unless there's a change I can't tell

mtgjson5/resources/additional_sets.json
```diff
@@ -0,0 +1,70 @@
+{
+    "dd3": {
```
ZeldaZach: CAPS please

mtgjson5/resources/additional_sets.json
```diff
@@ -0,0 +1,70 @@
...
+        "digital":false,
+        "nonfoil_only":false,
+        "foil_only":false,
+        "search_uri":"https://api.scryfall.com/cards/search?include_extras=true\u0026include_variations=true\u0026order=set\u0026q=e%3Aq04\u0026unique=prints"
+    },
+    "q05": {
+        "name":"Challenger Decks 2021",
+        "code":"q05",
+        "set_type":"box",
+        "icon_svg_uri":"https://svgs.scryfall.io/sets/default.svg?1688356800",
```
ZeldaZach: I wouldn't include a cache buster ID, just the ".svg" should be enough

mtgjson5/set_builder.py
axxroytovu: Fixed

mtgjson5/__main__.py
axxroytovu: I thought about building a provider but we don't use providers for any other resource file.

mtgjson5/__main__.py
axxroytovu: Fixed

mtgjson5/providers/cardkingdom.py
axxroytovu: Yeah this was a tox change and I didn't question it. I can remove these if you want.

mtgjson5/resources/additional_sets.json
axxroytovu: Fixed

mtgjson5/resources/additional_sets.json
axxroytovu: Ok. The scryfall api provides the cache ID so I just kept it.

mtgjson5/__main__.py
ZeldaZach: I don't think we need a provider, more so a common util method to get access to this data

mtgjson5/set_builder.py
```diff
@@ -404,7 +404,13 @@ def build_mtgjson_set(set_code: str) -> Optional[MtgjsonSetObject]:
     # Ensure we have a header for this set
     set_data = get_scryfall_set_data(set_code)
     if not set_data:
-        return None
+        with constants.RESOURCE_PATH.joinpath("additional_sets.json").open(
+            encoding="utf-8"
+        ) as f:
+            additional_sets_data = json.load(f)
+            set_data = additional_sets_data.get(set_code.upper(), False)
```
ZeldaZach:
```suggestion
            set_data = additional_sets_data.get(set_code.upper())
```

mtgjson5/set_builder.py
```diff
@@ -404,7 +404,13 @@ def build_mtgjson_set(set_code: str) -> Optional[MtgjsonSetObject]:
...
+        if not set_data:
```
ZeldaZach, line 409: I don't see a need for this if statement tbh, as set_data is either set or it's None

axxroytovu: If set_data is None, then it needs to close the function and not try building the rest of the data with no inputs. Line 406 checks if the set code was valid in Scryfall, and line 412 checks if the set code is valid in the local file. If the set code is in neither repository, it returns None.

mtgjson5/__main__.py
axxroytovu: Reworked to use util function

PR #1117, opened 2023-09-11
preserve commander property per card

kodawah:
In this way it's possible to tag if a card is the commander card of the deck.


mtgjson5/classes/mtgjson_set_deck.py
```diff
@@ -21,11 +21,13 @@ class MtgjsonSetDeckCardObject:
         uuid: str
         count: int
         finish: str
+        tags: Optional[List[str]]
 
-        def __init__(self, uuid: str, count: int, finish: str) -> None:
+        def __init__(self, uuid: str, count: int, finish: str, tags: List[str]) -> None:
```
ZeldaZach, line 26: tags should be Optional in the param

mtgjson5/providers/cardkingdom.py
```diff
@@ -60,7 +60,8 @@ def strip_sealed_name(self, product_name: str) -> str:
         """
         Cleans and strips sealed product names for easier comparison.
         """
-        name = re.sub(r"[^\w\s]", "", product_name)
+        name = product_name.strip()
```
ZeldaZach, line 63: can one line this below

ZeldaZach:
Superseded by #1118

PR #1118, opened 2023-09-11
Fix #1114: Copy structure of Decks/ to AllPrintings/

### LINKED ISSUE #1114: for decklists, preserve the commander property
When loading up decklists, it would be nice to preserve the `commander` property of the card

The json could be modified to host a series of extra tags

```
{
count: 1
uuid: $uuid
tags: ["commander"]
}
```

ZeldaZach:
Instead of using a new struct inside of set.decks, use the same format as the decks/ dir, just omit information that can be determined
[DIS.json.txt](https://github.com/mtgjson/mtgjson/files/12572421/DIS.json.txt)
[C17.json.txt](https://github.com/mtgjson/mtgjson/files/12572439/C17.json.txt)


kodawah:
should be filtering out the `"isFoil": false,` as they are unneeded

also isn't it possible to keep using "foil" as key? it introduces breakage for no gain

finally, it would be nice to leave the option open to implement #1114

mtgjson5/classes/mtgjson_card.py
```diff
@@ -57,6 +57,7 @@ class MtgjsonCardObject:
     has_non_foil: Optional[bool]  # Deprecated - Remove in 5.3.0
     identifiers: MtgjsonIdentifiersObject
     is_alternative: Optional[bool]
+    is_foil: Optional[bool]
```
kodawah, line 60: what about `is_etched`?

kodawah: to elaborate, since mtgjson/scryfall conflate the finishes, it's not possible to have a flag that says whether a card is foil or not at the base level. What I think the intent was `is_foil_only` which can be derived from `len(finishes) == 1 && finishes[0] == "foil"` - a oneliner is not worth a new variable IMO

ZeldaZach: This variable is only used for the in-line decks, and it's a legacy field for that at this time. We can look to expand upon it in the future, sure, but I'd like to keep some sense of consistency

ZeldaZach:
> should be filtering out the "isFoil": false, as they are unneeded
This is defined in the decks/ impl, which has it as false/true, so I'd leave it as-is here

kodawah:
>  This is defined in the decks/ impl, which has it as false/true, so I'd leave it as-is here

so what? it's a json standard that lack of a boolean means false, it's effectively redundant data

mtgjson5/classes/mtgjson_card.py
kodawah: if you want consistency, drop `isFoil`, use `finish` and mark it as a string

PR #1151, opened 2023-11-10
Manually re-add set name translations

### LINKED ISSUE #1056: Translations property of every Set is empty
MTGJSON 5.2.1+20230409

#### Description of Bug
All Sets have an empty "translations" field, it is shown like empty object "{}".

I have tried several times over the last month with this file: https://mtgjson.com/api/v5/SetList.json
But it happens in all files that the definition of "Set" is included.

ZeldaZach:
- Fix #1056


kodawah (approved):
code is ok but it's a maintenance burden

PR #1160, opened 2023-12-15
Add Scryfall card_back_id to card identifiers

gwax:
The card_back_id for Scryfall is an identifier for the image on the back of a physical card. This is distinct from the image for the card on the back of a DFC. Specifically, in the case of Meld cards, the card_back_id can be used to fetch the image for the half of the rear card that is on the given front face.

Take, for example, BRO - Argoth, Sanctum of Nature The front of the card has scryfall id = b29c9e4f-7b98-4610-a681-ae6297e8fc72 The front has illustration id = 5e618b24-3787-4e98-bccf-41059a1f6406 The front has card_back_id = 17471e86-f808-418a-9ac2-ce0ab092afbd The back has scryfall id = 414b9230-9d25-4bdf-8b1e-b4fa2035b6a4 The back has illustration id = a72a2bea-1706-458c-a20d-7ab4a8d3ccf0

The image on the back (lower half of Titania, Gaea Incarnate) is only accessible via card_back_id as:
https://backs.scryfall.io/large/1/7/17471e86-f808-418a-9ac2-ce0ab092afbd.jpg

### DIFF
--- a/mtgjson5/classes/mtgjson_identifiers.py
+++ b/mtgjson5/classes/mtgjson_identifiers.py
@@ -26,6 +26,7 @@ class MtgjsonIdentifiersObject:
     multiverse_id: Optional[str]
     scryfall_id: Optional[str]
     scryfall_illustration_id: Optional[str]
+    scryfall_card_back_id: Optional[str]
     scryfall_oracle_id: Optional[str]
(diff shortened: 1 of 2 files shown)


gwax:
I'll need to change the branchname for the container build to work properly

PR #1199, opened 2024-05-03
Append sealed product ids to tcgplayer skus file

kodawah:
This PR adds the SKU ids of the sealed products to the TcgplayerSkus.json file


kodawah:
PR updated to a 1liner

PR #1208, opened 2024-06-30
Use new CardMarket API for Prices

ZeldaZach:
- Pt1 of Migration

### DIFF
--- a/mtgjson5/providers/cardmarket/monolith.py
+++ b/mtgjson5/providers/cardmarket/monolith.py
@@ -1,20 +1,15 @@
 """
 MKM 3rd party provider
 """
-import base64
-import io
 import json
 import logging
-import math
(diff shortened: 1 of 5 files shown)


mtgjson5/providers/cardmarket/monolith.py
```diff
@@ -36,6 +31,7 @@ class CardMarketProvider(AbstractProvider):
 
     connection: Mkm
     set_map: Dict[str, Dict[str, Any]]
+    PRICE_GUIDE_URL: str = "https://downloads.s3.cardmarket.com/productCatalog/priceGuide/price_guide_1.json"
```
ZeldaZach: Note to self: Probably worth moving to a config in case the URL changes over time, easier to manage

PR #1239, opened 2024-11-13
Add FWB as an empty set to capture sealed product

axxroytovu:
Foreign White Border products are typically tracked separately from the Foreign Black Border products and should have their own page.

https://www.cardmarket.com/en/Magic/Expansions/Foreign-White-Bordered

### DIFF
--- a/mtgjson5/resources/additional_sets.json
+++ b/mtgjson5/resources/additional_sets.json
@@ -88,5 +88,18 @@
         "nonfoil_only": true,
         "foil_only": false,
         "search_uri": "https://api.scryfall.com/cards/search?include_extras=true&include_variations=true&order=set&q=e%3Amb1&unique=prints"
+    },
+    "FWB": {
+        "name": "Foreign White Border",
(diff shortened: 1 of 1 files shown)


mtgjson5/resources/additional_sets.json
```diff
@@ -88,5 +88,18 @@
...
+    "FWB": {
+        "name": "Foreign White Border",
+        "code": "FWB",
+        "released_at": "1994-04-11",
+        "set_type": "core",
+        "parent_set_code": "3ed",
+        "digital": false,
+        "nonfoil_only": true,
+        "foil_only": false,
+        "block_code": "lea",
```
kodawah, line 101: if parent code is `3ed`, is it correc that the block code is `lea`? should it not be `fbb` since it contains all the fbb cards just in white border?

ZeldaZach:
Is this still needed? If so, can we get a rebase and I'll merge it in!

kodawah:
(yes it's still needed)

PR #1245, opened 2025-02-17
chore: add mkm name fixes for sld

Emnalyeriar:
Description:
I've added mappings for MKM Secret Lair Drop Series set names to their corresponding MTGJSON set names in `mkm_set_name_fixes.json`. 

Changes made:
- Added new mappings for SLD sets in the name fixes file
- This is my first contribution to this mapping, so I would appreciate if someone could review:
  - The correctness of the mappings
  - The JSON formatting
  - The overall approach to handling SLD set name mappings

I'm not entirely sure if this is the correct way to handle these mappings, so any feedback or guidance would be greatly appreciated. Please let me know if any adjustments are needed.

Thanks for reviewing!


ZeldaZach:
Mapping looks correct, JSON Formatting is correct.

As for approach, I think it's okay for now. Maybe we'll want an override rule in the code itself to just look for "secret lair" in the name, but for now this is good to backfil until I can do that.

PR #1277, opened 2025-07-29
Support vehicles and spacecraft as commander

j7126:
As of Edge of Eternities, legendary vehicles and spacecraft with a power/toughness box can now be a commander. This change updates the commander leadership skill as such.


ZeldaZach:
Thanks for your contribution towards making MTGJSON a better service :)

PR #1304, opened 2025-09-18
Update MTGJSON UUID Generator + UUID Cache

ZeldaZach:
- Change how MTGJSONv5 UUIDs are generated, using SF_ID+SIDE instead of all of the other attributes
- Cache a local copy of UUIDs generated prior to this change, and use those values instead of the newly generated ones
- Improves stability (+ trust) in our UUID system


kodawah (approved):

danbopes:
I understand `_name` is nice to have from a debug/Ctrl+f point of view, but from a purely optimization point of view, that's a lot of extra data we're throwing into memory and needing to read/parse through which ultimately isn't ever used. I'd distribute a super minimal json to the clients, with maybe a link to a GitHub blob/hard link to a blob with the debug fields.

mtgjson5/set_builder.py
```diff
@@ -713,59 +748,26 @@ def add_uuid(
...
-                )
-            )
+    # MTGJSONv5 UUIDs will now be generated using
+    # Scryfall_UUID + Side ("a" by default if side not specified).
+    # For UUIDs that have been generated in the past, we will continue
+    # to honor those and, instead of regenerating them, will load them
+    # from a backup cache
+    cached_mtgjson_uuid = UuidCacheProvider().get_uuid(
+        str(mtgjson_object.identifiers.scryfall_id), (mtgjson_object.side or "a")
+    )
```
brentmclark, line 762: Is there any value in moving the logic that defaults to side `a` into the `get_uuid` method so future call sites won't have to implement the same null coalescing logic and/or remember the correct default?

brentmclark (commented):
Shifting all the v4 uuid logic into a method in set_builder was a positive change IMO.  The code is easier to read and reason about.  I also appreciate the switch to early returns instead of the nested conditions.

mtgjson5/set_builder.py
ZeldaZach: Maybe, but it's used in two places (a few lines below) so it's not a huge deal. Also the goal is for this to be a one off operation

PR #1312, opened 2025-09-22
Tcg 79/enhance tcgplayer pricing fields

genericlady:
1. Complete pricing data capture from TCGPlayer API
2. Parity with enhanced pricing model similar to TCGCSV integration
3. No breaking changes for existing MTGJSON consumers
4. Production ready - just needs TCGPlayer API credentials to activate

### DIFF
--- a/mtgjson5/classes/mtgjson_prices.py
+++ b/mtgjson5/classes/mtgjson_prices.py
@@ -23,6 +23,20 @@ class MtgjsonPricesObject(JsonObject):
     sell_normal: Optional[float]
     sell_foil: Optional[float]
     sell_etched: Optional[float]
+    
+    # Enhanced TCGPlayer pricing fields
+    sell_normal_low: Optional[float]
+    sell_normal_mid: Optional[float]
+    sell_normal_high: Optional[float]
+    sell_normal_direct_low: Optional[float]
+    sell_foil_low: Optional[float]
+    sell_foil_mid: Optional[float]
+    sell_foil_high: Optional[float]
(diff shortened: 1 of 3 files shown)


mtgjson5/classes/mtgjson_prices.py
```diff
@@ -85,5 +126,45 @@ def to_json(self) -> Dict[str, Any]:
...
+        
+        # Add enhanced pricing fields if any are present
+        has_enhanced_pricing = any([
+            self.sell_normal_low, self.sell_normal_mid, self.sell_normal_high, self.sell_normal_direct_low,
+            self.sell_foil_low, self.sell_foil_mid, self.sell_foil_high, self.sell_foil_direct_low,
+            self.sell_etched_low, self.sell_etched_mid, self.sell_etched_high, self.sell_etched_direct_low
+        ])
+        
+        if has_enhanced_pricing:
+            buy_sell_option["retail_enhanced"] = defaultdict(lambda: defaultdict(dict))
```
genericlady, line 138: good idea for extendable interface
setup of raw data model transforming out of it's respective scope what it was going to look like
the model is no longer 1:1 with json output

having a retail enhanced object with provider context
new model with default interface that we can work around with it's own to_json

```
prices_tcgplayer = {
 low, mid, high
}
manapool = {
sell normal near mint high whatever it is
}
```

per provider level

current output has by_price and sell_price add a new key into the dictionary that key would include the retail specific pricing content. don't want to break current tenants.

PR #1316, opened 2025-09-29
Export the sourceSetCodes list for decks

kodawah:
this will unlock https://github.com/mtgjson/mtg-sealed-content/pull/361/files
(thanks to @taw for guidance on this)


kodawah:
updated to fix tox and add field to the definition

ZeldaZach (approved):

PR #1319, opened 2025-10-05
TCG-110: Implement AllPrices v2 schema for MTGJSON price data compilation

genericlady:
## Summary

Implements AllPrices v2 schema to produce normalized price outputs (AllPricesv2.json, AllPricesTodayv2.json) alongside existing AllPrices files. Downstream consumers can now opt into the richer dataset without disruption.

Each provider already gathers card prices and hands them to the PriceBuilder. Right now they only send one price per card finish. We’ll teach them to send every price they know about instead. Using TCGplayer as the example: their API gives us several numbers for each card—market, low, mid, high, and TCG Direct low, plus the buylist prices. We’ll update the TCGplayer provider so, when it talks to PriceBuilder, it includes all those numbers. The builder then turns each number into its own record in the new price format, so the final file keeps every price point instead of just one.

### New V2 Price Record Container
- Added `MtgjsonPricesRecordV2` dataclass with fields: provider, treatment, currency, price_value, price_variant, uuid, platform, price_type, date, and optional subtype
- Added `MtgjsonPricesV2Container` class for organizing records by provider with JSON serialization to `{provider: [records...]}` format

### PriceBuilder Updates
- Extended `build_prices()` to return both legacy and v2 formats: `((all_prices, today_prices), (v2_all, v2_today))`
- Added `convert_legacy_to_v2()` static method to transform legacy nested dict format to v2 normalized records
- Maintained backward compatibility - legacy tuple remains unchanged

### Output Generator Extensions
- Updated `generate_compiled_prices_output()` to accept and write v2 structures
- Writes AllPricesv2.json and AllPricesTodayv2.json in parallel with legacy files
- Preserved existing write semantics

### Structure Object Updates
- Added `all_prices_v2` and `all_prices_today_v2` properties to `MtgjsonStructuresObject`
- New files automatically included in CompiledList metadata

### Testing
- Added `test_convert_legacy_to_v2()` to verify legacy-to-v2 conversion
- Added `test_v2_record_to_json()` and `test_v2_record_to_json_without_subtype()` for serialization
- All 8 tests pass successfully

## Testing Performed

```bash
source venv/bin/activate && python -m pytest tests/mtgjson5/test_today_price_builder.py -v
# 8 passed in 0.16s
```

Linters run:
- black (formatted)
- All existing tests still pass

## Schema Documentation

Inline docstrings explain:
- V2 vs legacy format differences
- Field meanings and usage
- Serialization contract

## Implementation Notes

- No breaking changes to existing API
- Memory overhead is minimal - v2 conversion happens after legacy processing
- Schema validation baked into dataclass structure
- V2 files automatically included in compiled output metadata


mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,141 @@
...
+            "date": self.date,
+        }
+
+        if self.subtype is not None:
+            result["subtype"] = self.subtype
+
+        return result
+
+
+class MtgjsonPricesV2Container:
```
ZeldaZach: Not a huge fan of having multiple classes inside the same file necessarily. Does it make any sense to split these up?

mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,141 @@
...
+    Container for organizing v2 price records by provider.
+
+    This class manages the aggregation of price records and provides
+    serialization to the {provider: [records...]} format expected
+    by the AllPricesv2.json output.
+    """
+
+    def __init__(self) -> None:
+        """Initialize an empty price records container."""
+        self._records: Dict[str, List[MtgjsonPricesRecordV2]] = {}
```
ZeldaZach: I prefer having class variables defined in the "header" of the class
```
class Foo:
   var: type
```

mtgjson5/price_builder.py
```diff
@@ -187,6 +188,64 @@ def write_price_archive_data(
             f"Finished compressing content to {local_save_path} (Size = {local_save_path.stat().st_size} bytes)"
         )
 
+    @staticmethod
+    def convert_legacy_to_v2(legacy_prices: Dict[str, Any]) -> MtgjsonPricesV2Container:
```
ZeldaZach: Is this considered a one time conversion operation? Do we want to start AllPricesv2 as a new clean slate or backfill it? Just curious on your ideas

ZeldaZach (commented):
The general setup here I like, but I don't see a proof of concept for any provider on how the new data would look. If we imported TCGPlayer, what would those changes look like? Is there a fast follow up to include at least one provider that would directly write today's prices in the new v2 format?

mtgjson5/classes/mtgjson_prices_v2.py
genericlady: good catch

mtgjson5/price_builder.py
genericlady: Yeah lets discuss tonight what we should do here. If we backfill it then people can start wiring it in but I'm also ok with not doing that. I think it can make things confusing.

I think I'm leaning towards not doing this.

genericlady: bleh i think i'm just gonna remove it lol

mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,78 @@
...
+    provider: str
+    treatment: str
+    currency: str
+    price_value: float
+    price_variant: str
+    uuid: str
+    platform: str
+    price_type: str
+    date: str
+    subtype: Optional[str] = None
```
ZeldaZach, line 52: Define optional values in the `__init__`

mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,78 @@
...
+    - uuid: MTGJSON card UUID this price applies to
+    - platform: Trading platform ('paper', 'mtgo')
+    - price_type: Transaction type ('retail', 'buy_list')
+    - date: ISO date string for when this price was recorded
+    """
+
+    provider: str
+    treatment: str
+    currency: str
+    price_value: float
```
ZeldaZach, line 46: price instead?

mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,78 @@
...
+    price_variant: str
```
ZeldaZach, line 47: Thinking about the name on this

mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,78 @@
...
+    def to_json(self) -> Dict[str, Any]:
+        """
+        Serialize the price record to JSON format.
+
+        Returns a dictionary with camelCase keys for JSON output.
+        Excludes None values for optional fields.
+
+        :return: JSON-serializable dictionary
+        """
+        result = {
```
ZeldaZach, line 63: There's a better way to iterate the class. Look at other example classes to_json() methods

mtgjson5/classes/mtgjson_prices_v2_container.py
```diff
@@ -0,0 +1,99 @@
...
+
+This module defines the container class for organizing v2 price records by provider.
+"""
+
+from typing import Any, Dict, List
+
+from .mtgjson_prices_v2 import MtgjsonPricesRecordV2
+
+
+class MtgjsonPricesV2Container:
```
ZeldaZach, line 12: I'm not certain about this class/container. A List of MTGJSONPricesV2 might be sufficient? Or a Dict of that?

mtgjson5/compiled_classes/mtgjson_structures.py
```diff
@@ -58,6 +59,8 @@ def __init__(self) -> None:
         self.atomic_cards = "AtomicCards"
         self.all_prices = "AllPrices"
         self.all_prices_today = "AllPricesToday"
+        self.all_prices_v2 = "AllPricesv2"
```
ZeldaZach, line 62: Should we do capital V? @kodawah

mtgjson5/output_generator.py
```diff
@@ -31,15 +31,26 @@
 
 
 def generate_compiled_prices_output(
-    all_price_data: Dict[str, Any], today_price_data: Dict[str, Any], pretty_print: bool
+    all_price_data: Dict[str, Any],
+    today_price_data: Dict[str, Any],
+    all_price_data_v2: Any,
```
ZeldaZach, line 36: I'd like types defined over Any

mtgjson5/price_builder.py
```diff
@@ -113,6 +114,42 @@ def build_today_prices(self) -> Dict[str, Any]:
...
+        """
+        Build v2 price records by calling build_v2_prices() on each provider.
+
+        Aggregates v2 records from providers that support the v2 interface.
+        Providers without v2 support will return empty lists.
+
+        :return: Container with all v2 price records from today
+        """
+        if not self.all_printings_path.is_file():
+            LOGGER.error(
```
ZeldaZach, line 127: I think this is wrong, we should download the AllPrintings if it doesn't exist for price builds

mtgjson5/price_builder.py
```diff
@@ -113,6 +114,42 @@ def build_today_prices(self) -> Dict[str, Any]:
...
+                f"Unable to build v2 prices. AllPrintings not found in {MtgjsonConfig().output_path}"
+            )
+            return MtgjsonPricesV2Container()
+
+        v2_container = MtgjsonPricesV2Container()
+
+        for provider in self.providers:
+            provider_name = type(provider).__name__
```
ZeldaZach, line 135: this feels suspect at best. I don't like depending on double underscore components if not necessary

mtgjson5/price_builder.py
```diff
@@ -252,3 +297,159 @@ def build_prices(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+        if not MtgjsonConfig().has_section("Prices"):
+            return today_prices_v2, today_prices_v2
+
+        bucket_name = MtgjsonConfig().get("Prices", "bucket_name")
+        bucket_object_path_v2 = MtgjsonConfig().get(
+            "Prices", "bucket_object_path_v2", fallback="AllPricesv2.json.tar.xz"
+        )
+
+        # Download existing v2 archive (dict format: {provider: [records...]})
+        archive_dict = self.get_price_archive_data(bucket_name, bucket_object_path_v2)
```
ZeldaZach, line 333: This is a full re implementation of the current archive process for pricing. I'd rather abstract that code and call it in the two places vs a copy.

mtgjson5/price_builder.py
```diff
@@ -252,3 +297,159 @@ def build_prices(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+        self, archive_dict: Dict[str, Any]
+    ) -> MtgjsonPricesV2Container:
+        """
+        Reconstruct a MtgjsonPricesV2Container from archived dict format.
+        Archive format is {provider: [record_dicts...]}
+
+        :param archive_dict: Downloaded archive data
+        :return: Reconstructed container
+        """
+        from .classes.mtgjson_prices_v2 import MtgjsonPricesRecordV2
```
ZeldaZach, line 369: MTGJSON has all imports at top of file

mtgjson5/price_builder.py
```diff
@@ -252,3 +297,159 @@ def build_prices(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+
+        LOGGER.info("Uploading v2 price data")
+        MtgjsonS3Handler().upload_file(
+            str(local_zip_file_v2), bucket_name, bucket_object_path_v2
+        )
+        local_zip_file_v2.unlink()
+
+        return archive_prices_v2, today_prices_v2
+
+    def _reconstruct_v2_container_from_dict(
```
ZeldaZach, line 359: So this class feels a bit funky to me. I'd rather store the backup copy as pretty much a copy of what AllPricesV2 would look like as an output, and then append to it with new data daily. This seems to store as a JSON blob of components vs a final structure?

mtgjson5/price_builder.py
```diff
@@ -252,3 +297,159 @@ def build_prices(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+        container = MtgjsonPricesV2Container()
+        if not archive_dict:
+            return container
+
+        LOGGER.info("Loading existing v2 archive data")
+        # Archive is {provider: [records...]} format
+        for provider, records_list in archive_dict.items():
+            for record_dict in records_list:
+                # Reconstruct MtgjsonPricesRecordV2 from dict
+                record = MtgjsonPricesRecordV2(
```
ZeldaZach, line 380: This re-construction of an object doesn't feel very necessary. There's a few ways to do this, but I think we can skip this step.

mtgjson5/price_builder.py
```diff
@@ -252,3 +297,159 @@ def build_prices(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+        :param archive: Archive container to merge into
+        :param today: Today's prices to merge from
+        """
+        # Simply add all today's records to archive
+        # Each record is timestamped with today's date, so we preserve history
+        # pylint: disable=import-outside-toplevel
+        from .classes.mtgjson_prices_v2 import MtgjsonPricesRecordV2
+
+        today_serialized = today.to_json()
+        for _, records_list in today_serialized.items():
```
ZeldaZach, line 414: I like the general idea of the method, but again creating these records could be simpler

mtgjson5/price_builder.py
```diff
@@ -252,3 +297,159 @@ def build_prices(self) -> Tuple[Dict[str, Any], Dict[str, Any]]:
...
+        :param archive: Archive container to prune
+        """
+        LOGGER.info("Pruning v2 archive data")
+        cutoff_date = datetime.date.today() - datetime.timedelta(
+            weeks=int(MtgjsonConfig().get("Prices", "max_weeks", fallback="104"))
+        )
+        cutoff_date_str = cutoff_date.isoformat()
+
+        # Build a new container with only non-expired records
+        pruned_container = MtgjsonPricesV2Container()
```
ZeldaZach, line 445: This is an interesting way of pruning data, Would delete record make more sense vs the complete replacement?

mtgjson5/providers/abstract.py
```diff
@@ -158,3 +159,22 @@ def get_price_field_name(is_foil: bool, is_etched: bool, is_sell: bool) -> str:
         if is_foil:
             return "sell_foil" if is_sell else "buy_foil"
         return "sell_normal" if is_sell else "buy_normal"
+
+    def build_v2_prices(
+        self, all_printings_path: pathlib.Path  # pylint: disable=unused-argument
```
ZeldaZach, line 164: I assume this comment is temporary?

mtgjson5/providers/abstract.py
```diff
@@ -158,3 +159,22 @@ def get_price_field_name(is_foil: bool, is_etched: bool, is_sell: bool) -> str:
...
+        price variants (e.g., TCGPlayer's market/low/mid/high/directLow),
+        subtypes, and other metadata that would be lost in legacy conversion.
+
+        Default implementation returns empty list; providers supporting v2
+        should override this method.
+
+        :param all_printings_path: Path to AllPrintings.json for ID mapping
+        :return: List of v2 price records with full metadata
+        """
+        return []
```
ZeldaZach, line 180: This should be an abstract method probably

mtgjson5/providers/tcgplayer.py
```diff
@@ -565,3 +619,159 @@ def get_tcgplayer_prices_map(
...
+    records: List[MtgjsonPricesRecordV2] = []
+    today_date = TCGPlayerProvider().today_date
+
+    # Map price field names to price_variant values
+    price_variants = {
+        "marketPrice": "market",
+        "lowPrice": "low",
+        "midPrice": "mid",
+        "highPrice": "high",
+        "directLowPrice": "direct_low",
```
ZeldaZach, line 652: Not sure how I feel about the underscore vs what we do with other things. Might make more sense as a space or directLow

mtgjson5/providers/tcgplayer.py
```diff
@@ -565,3 +619,159 @@ def get_tcgplayer_prices_map(
...
+                    subtype = "Foil"
+
+            if treatment:
+                for uuid in keys:
+                    record = MtgjsonPricesRecordV2(
+                        provider="tcgplayer",
+                        treatment=treatment,
+                        currency="USD",
+                        price_value=float(high_price),
+                        price_variant="high",
```
ZeldaZach, line 768: shouldn't we have a list of price variants?

tests/mtgjson5/test_today_price_builder.py
```diff
@@ -242,3 +246,143 @@ def test_multiverse_bridge_cardsphere_build_today_prices():
...
+        uuid="test-uuid-123",
+        platform="paper",
+        price_type="retail",
+        date="2024-01-15",
+        subtype="premium",
+    )
+
+    result = record.to_json()
+
+    assert result["provider"] == "testprovider"
```
ZeldaZach, line 268: Instead of raw asserts, look into https://docs.python.org/3/library/unittest.html

tests/mtgjson5/test_today_price_builder.py
```diff
@@ -242,3 +246,143 @@ def test_multiverse_bridge_cardsphere_build_today_prices():
...
+def test_tcgplayer_build_v2_prices():
+    """Test TCGPlayerProvider.build_v2_prices() with all price variants."""
+    provider = TCGPlayerProvider()
+    patch.object(
+        provider,
+        "get_tcgplayer_magic_set_ids",
+        return_value=json.load(
+            get_resource_file_buffer("tcgplayer_magic_set_ids.json")
+        ),
+    ).start()
```
ZeldaZach, line 311: I have 0 idea what this is... comments? Or more details?

ZeldaZach (changes requested):
Thanks so much for your efforts!

With that said, this doesn't really do what I was hoping it would do. My general idea was to have a mtgjson_prices_v2 struct (like you have) and keep a running Dict of them. 

My original idea was to have a v2 creator, then a v1 downgrader which pulls data from v2 and creates the v1 versionset. I can understand your reasoning on how this would be safer, but maybe we will consider a change over when all providers are onboarded.

I'm also very hesitant on AI code that we might not understand... I'd like simpler code that's more readable vs necessarily the most efficient to get the job done.

cc: @kodawah Need some feedback on how we'd like to see V2 function

mtgjson5/classes/mtgjson_prices_v2.py
kodawah, line 46: any strong opinion on using floats? in general it's bad practice to use them for prices, and as we're adding a new file, we might be able to fix this small problem https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency

(however i see the argument that price representation should be left to the user and mtgjson should focus on clarity, so it's ok to ignore this point too)

mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    MTGJSON Prices Record V2 Object
+
+    Represents a single price data point with full context about provider,
+    treatment (foil/non-foil/etched), platform (paper/mtgo), and price type
+    (retail/buy_list).
+
+    Schema Fields:
+    - provider: Price provider name (e.g., 'cardkingdom', 'tcgplayer')
+    - treatment: Card finish type ('normal', 'foil', 'etched')
```
kodawah, line 29: if this field is only used for finish, we should name it `finish` and keep the enum value the same as upstream (`nonfoil`, `foil`, `etched`)

mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,78 @@
...
+    - subtype: Optional additional categorization (e.g., price tier)
```
kodawah, line 30: not sure what this field is for

mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,78 @@
...
+    - currency: Currency code (e.g., 'USD', 'EUR')
+    - price_value: The actual price value as a float
+    - price_variant: Provider-specific price tier/type:
+        * 'legacy': Used for conversions from legacy format where variant is unknown
+        * TCGPlayer: 'market', 'low', 'mid', 'high', 'tcgdirect_low'
+        * Other providers may have their own variants
+    - uuid: MTGJSON card UUID this price applies to
+    - platform: Trading platform ('paper', 'mtgo')
```
kodawah, line 38: we could skip this and move mtgo to the currency as tix

mtgjson5/classes/mtgjson_prices_v2.py
```diff
@@ -0,0 +1,78 @@
...
+    - price_type: Transaction type ('retail', 'buy_list')
```
kodawah, line 39: since it's the side of the same store, how's the name `side` for this field? also no "_" in `buylist` 🙏

mtgjson5/compiled_classes/mtgjson_structures.py
kodawah: yes

mtgjson5/providers/tcgplayer.py
```diff
@@ -266,6 +270,56 @@ def generate_today_price_dict(
...
+            build_tcgplayer_v2_retail_records,
+            ids_and_names,
+            repeatable_args=[
+                tcg_foil_and_non_foil_to_mtgjson_map,
+                tcg_etched_foil_to_mtgjson_map,
+            ],
+            fold_list=True,
+        )
+
+        LOGGER.info("Building TCGPlayer v2 buylist data")
```
kodawah, line 308: tcgplayer doesn't have a buylist any more to my knowledge, did things change?

PR #1325, opened 2025-10-08
Migrate project management to `uv`

### LINKED ISSUE #1324: Migrate to modern python project management tools (uv)
Use `uv` and `pyptoject.toml` are encouraged by the [python packaging guide](https://packaging.python.org/en/latest).
[`uv`](https://docs.astral.sh/uv/) is written in Rust, blazing fast, and can, among other things:
- manage the local python version used
- manage the runtime virtual environment
- solve & lock dependencies versions
- publish to indexes like pypi.org

liberforce:
[uv](https://docs.astral.sh/uv/) is a python project manager. It's fast, handles solving dependencies, can install and manage multiple python versions and builds. It supports Windows, macOS and Linux.

Fixes #1324


liberforce:
Still a few problems:
- I need to figure out how to properly handle the version number in pyproject.toml, since it was modified on the fly using mtgjson.properties
- I've generated requirement files using uv, but just because they are used by the docker image and the CI. They should not be required.

ZeldaZach:
Hey there! I have a few questions regarding your changes.

While I'm generally not a fan of changing our build pipelines as a first commit from a new author, I'm definitely interested in what you have to contribute.

That being said, why are there so many new requirements? That's a bit of a concern for me as I don't recognize most of them and locking (potential?) transitive dependencies doesn't seem correct to me. You've also updated the requirements_test to include everything necessary for dev work, which is probably correct(?), but I've always just installed both files for dev work.

You've also pointed out that the version numbers are tricky to handle. Maybe we can work on that still. Also, does this impact our build pipeline in AWS at all? Will our cloud deployment need updates to support such a change? The AWS code is private, but if changes are needed I'd like to get those in before things break

ZeldaZach:
Superseded by #1445 -- Thanks so much for introducing me to uv, this looks really useful!

PR #1329, opened 2025-10-16
Fix possible call of method on `None`

### LINKED ISSUE #1328: Possible call of method on None
MTGJSON 5.2.2+20251015"

#### Description of Bug

By updating the development dependencies using my `uv` branch, mypy started throwing this error:

```
mtgjson5/providers/scryfall/orientation_detector.py:52: error: Item "None" of "Tag | None" has no attribute "get"  [union-attr]
mtgjson5/providers/scryfall/orientation_detector.py:60: error: List comprehension has incompatible type List[str | AttributeValueList | None]; expected List[str]  [misc]
```

Here's the relevant link: https://github.com/mtgjson/mtgjson/blob/a79eb46efd2dc24729122863e6ecc4f4d0b7da43/mtgjson5/providers/scryfall/orientation_detector.py#L48-L60

On line 52, there's a call to `bs4.Tag.find()` followed by a call to `get` on the returned value.
On line 59, there's a call to `bs4.Tag.find_all`, followed by a call to `get` on the returned value.

Both `find` and `find_all` may return `None` according to the documentation, which would make the call to `get` fail.

liberforce:
Fixes #1328

### DIFF
--- a/mtgjson5/providers/scryfall/orientation_detector.py
+++ b/mtgjson5/providers/scryfall/orientation_detector.py
@@ -49,12 +49,13 @@ class ScryfallProviderOrientationDetector(AbstractProvider):
         """
         Parse card orientation from bs4 Tag
         """
-        return str(orientation_header.find("a").get("id"))
+        a_tags = orientation_header.find("a") or {}
+        return str(a_tags.get("id"))
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


ZeldaZach (approved):
Thanks for the fix!

PR #1331, opened 2025-10-17
Scryfall rate limit is 10 req/s

### LINKED ISSUE #1330: mtgjson doesn't respect Scryfall's rate limiting guidelines
MTGJSON 5.2.2+20251017

#### Description of Bug

The scryfall provider currently uses a 15 req/s rate limiting. However, per [Scryfall rate limiting rules](https://scryfall.com/docs/api#rate-limits-and-good-citizenship), we should max at 10 req/s.

By just changing this parameter, we get a 3.3× speedup building the Onslaught (ONS) set:

```
cache(disabled), rate-limit(15 req/s): 9min 43s
cache(disabled), rate-limit(10 req/s): 2min 58s
```

liberforce:
See https://scryfall.com/docs/api#rate-limits-and-good-citizenship

Fixes #1330

### DIFF
--- a/mtgjson5/providers/scryfall/monolith.py
+++ b/mtgjson5/providers/scryfall/monolith.py
@@ -97,7 +97,7 @@ class ScryfallProvider(AbstractProvider):
         return all_cards
 
     @ratelimit.sleep_and_retry
-    @ratelimit.limits(calls=15, period=1)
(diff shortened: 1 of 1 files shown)


ZeldaZach:
Pending on #1333 to see if this is a problem with our build chain

PR #1333, opened 2025-10-20
feat(scryfall): warn once we reach API's rate limit

### LINKED ISSUE #1332: Warn when scryfall rate limit is reached
Hitting the rate limit hugely impacts performance (3min→10min when building the ONS set during my tests). Depending on the API key used, the rate limit can be different too. Adding warnings should make it easier to see when we hit the limit.

liberforce:
Fixes #1332


ZeldaZach:
Can you update your PR to pass the builds? Thanks!

PR #1338, opened 2025-10-30
Tcgplayer price points migration to v2 add tests

genericlady:
Summary

Successfully implemented test coverage for get_tcgplayer_prices_map() using VCR.py for HTTP recording without live network calls.

Changes Made:

1. Added vcrpy dependency (requirements_test.txt)
◦  Version 7.0.0 for recording/replaying HTTP interactions
2. Created VCR configuration (tests/conftest.py)
◦  Filters sensitive data: Authorization headers, client_id, client_secret
◦  Configurable record modes (once/all/none)
◦  Matches on method, scheme, host, port, path, query
3. Created test file (tests/mtgjson5/providers/test_tcgplayer_prices_map_shape.py)
◦  Tests return type: dict[str, MtgjsonPricesObject]
◦  Validates metadata: source="paper", provider="tcgplayer", currency="USD"
◦  Verifies field mapping: Normal→sell_normal, Foil→sell_foil, Etched→sell_etched
◦  Confirms no buylist fields set (function only handles retail prices)
◦  Tests empty results handling
4. Created VCR cassettes (in tests/mtgjson5/providers/cassettes/)
◦  tcgplayer_prices_map.yaml - normal test case with sample data
◦  tcgplayer_prices_map_empty.yaml - edge case with empty results
◦  All credentials filtered (no secrets committed)
5. Added documentation (tests/mtgjson5/providers/README_CASSETTES.md)
◦  Instructions for re-recording cassettes locally
◦  Security verification steps
◦  CI/CD guidance

Test Results:
•  ✅ All 3 tests pass
•  ✅ No network calls (verified with VCR_RECORD_MODE=none)
•  ✅ No credentials in cassettes (verified with grep)
•  ✅ Code formatted with black and isort


docs/plan
```diff
@@ -0,0 +1,57 @@
+Step 1 — Test the prices_map shape for TCGplayer (planning only)
```
genericlady, line 1: when we talk we should talk about what this is and why its here and if we're going to keep it. if we keep it then it should make sense and be numbered.

docs/tests/http_recording
```diff
@@ -0,0 +1,51 @@
+HTTP Recording with VCR.py (local re-recording)
```
genericlady, line 1: starting some docs. better late than never

tests/mtgjson5/providers/cassettes/tcgplayer_prices_map.yaml
```diff
@@ -0,0 +1,61 @@
+version: 1
```
genericlady: i guess it's a normal convention to keep the cassettes nested under the relevant tests? i wonder if it can be ./tests/cassettes ?

tests/mtgjson5/providers/README_CASSETTES.md
```diff
@@ -0,0 +1,80 @@
+# VCR Cassettes for TCGplayer Tests
```
genericlady: i wonder if the AI did this or vcr bootstrapped it. it should go into docs/tests/cassettes.md

tests/conftest.py
```diff
@@ -0,0 +1,55 @@
+"""
```
genericlady, line 1: This looks like it's part config and part helper for VCR. I guess this is fine? We need to take a closer look at vcr documentation

tests/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+Pytest configuration for MTGJSON tests.
+"""
+
+import os
+
+import pytest
+import vcr
+
+
+def filter_request(request):
```
genericlady: missing type annotation

tests/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+    if request.body:
+        # Filter client_id and client_secret from POST bodies
+        body_str = (
+            request.body.decode("utf-8")
+            if isinstance(request.body, bytes)
+            else request.body
+        )
+        if "client_id" in body_str or "client_secret" in body_str:
+            # Replace values but keep structure for replay
+            import re
```
genericlady: prefer import at the top of the file

docs/best_practices.md
```diff
@@ -0,0 +1,18 @@
+# Use Linter
```
genericlady, line 1: AI just did this and I might be ok with this. we need to start somewhere

PR #1347, opened 2025-11-26
Add Scryfall bulk models to Scryfall provider

the-muppet:
introduced pydantic models to validate and access bulk data and content from Scryfall


mtgjson5/providers/scryfall/monolith.py
```diff
@@ -8,10 +8,15 @@
 import re
 import sys
 import time
-from typing import Any, Dict, List, Optional, Set, Union
+from datetime import datetime
```
ZeldaZach: I prefer to use `import X` when possible, with the exception of typing and singleton

mtgjson5/providers/scryfall/monolith.py
```diff
@@ -22,6 +27,83 @@
...
+
+    def download(
+        self, session: Optional[requests.Session] = None, timeout: int = 300
+    ) -> List[Dict[str, Any]]:
+        """Download the bulk data file.
+
+        Uses a plain requests.Session to avoid caching large bulk files
+        that can exceed SQLite's blob size limits.
+        """
+        download_session = requests.Session()
```
ZeldaZach: I see you're using the requests session instead of the retryable_seession.Session() we have defined in the code (which takes advantage of the local caching opportunity). Was this intentional?

mtgjson5/providers/scryfall/monolith.py
```diff
@@ -22,6 +27,83 @@
...
+        if session:
+            download_session.headers.update(session.headers)
```
ZeldaZach: If you have a session, why are you still creating a new one and just copying the headers over? Feels a bit weird

mtgjson5/providers/scryfall/monolith.py
```diff
@@ -22,6 +27,83 @@
...
+        LOGGER.info(
+            f"Downloading bulk data: {self.name} ({self.size / 1_000_000:.1f}MB)"
+        )
+        response = download_session.get(str(self.download_uri), timeout=timeout)
+        response.raise_for_status()
+        result: List[Dict[str, Any]] = response.json()
```
ZeldaZach: I don't think you need to store as a temp var if you're just returning it as-is

mtgjson5/providers/scryfall/monolith.py
```diff
@@ -22,6 +27,83 @@
...
+    data: List[BulkDataObject] = Field(description="List of available bulk data files")
+
+    @classmethod
+    def fetch(
+        cls,
+        session: Optional[requests.Session] = None,
+        url: str = "https://api.scryfall.com/bulk-data/",
+    ) -> "BulkDataResponse":
+        """Fetch the bulk data index from Scryfall"""
+        session = session or requests.Session()
```
ZeldaZach: Again, we might want to take advantage of the cached session setup

mtgjson5/providers/scryfall/monolith.py
```diff
@@ -22,6 +27,83 @@
...
+        response = session.get(url, timeout=30)
```
ZeldaZach: timeout should be a var in fetch maybe, with default to 30?

mtgjson5/providers/scryfall/monolith.py
```diff
@@ -22,6 +27,83 @@
 LOGGER = logging.getLogger(__name__)
 
 
+BulkDataType = Literal[
```
ZeldaZach: Instead of adding to the monolith, maybe make a new file inside the SF dir for bulk work?

mtgjson5/providers/scryfall/monolith.py
```diff
@@ -273,3 +355,46 @@ def __get_card_names(self, url: str) -> List[str]:
         :return All unique card names found
         """
         return list({card["name"] for card in self.download(url).get("data", {})})
+
+
+    def get_bulk_cards(self, force_refresh: bool = False) -> pl.DataFrame:
+        cache_dir = pathlib.Path(MtgjsonConfig().get("MTGJSON", "cache_path"))
+        parquet_path = cache_dir / "all_cards.parquet"
```
ZeldaZach: I prefer .joinpath

mtgjson5/providers/scryfall/monolith.py
```diff
@@ -273,3 +355,46 @@ def __get_card_names(self, url: str) -> List[str]:
...
+            data = bulk_cards_object.download()
+            df = pl.DataFrame(data, infer_schema_length=None)
+            df.write_parquet(parquet_path)
+            return df
+
+        return pl.read_parquet(parquet_path)
+
+    def get_bulk_rulings(self, force_refresh: bool = False) -> pl.DataFrame:
+        cache_dir = pathlib.Path(MtgjsonConfig().get("MTGJSON", "cache_path"))
+        parquet_path = cache_dir / "rulings.parquet"
```
ZeldaZach: I prefer .joinpath

ZeldaZach (changes requested):

mtgjson5/providers/scryfall/monolith.py
mtg-elmo: Yes - as the 2.6-3GB file size of the bulk objects exceeds the built in SQLite capacity

ZeldaZach: Thanks for the comment on this, makes sense!

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
+"""
+Global cache for MTGJSON provider data and pre-computed aggregations.
+"""
+
+import json
+import logging
+from pathlib import Path
```
ZeldaZach: Please use the `import X` approach for all non typing/mtgjson5 stuff

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+        face.get("printed_name") or face.get("name", "")
+        for face in card_faces
+    )
+    
+    # Get face_name from the specific face
+    target_face = card_faces[face_idx]
+    face_name = target_face.get("printed_name") or target_face.get("name")
+    
+    # Build resolved entry
+    result = {
```
ZeldaZach, line 200: You don't have to store this as a temp var since we're returning it right away

ZeldaZach: Also, is this a Mtgjson Object of some kind? Maybe build it here?

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+        "text": target_face.get("printed_text") or entry.get("text"),
+        "type": target_face.get("printed_type_line") or entry.get("type"),
+        "flavor_text": target_face.get("flavor_text") or entry.get("flavor_text"),
+    }
+    
+    return result
+
+@cache
+def load_json(filename: str) -> Any:
+    return json.loads((constants.RESOURCE_PATH / filename).read_text())
```
ZeldaZach: I prefer joinpath over /.

This also has no validation filename is valid

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+
+
+class GlobalCache:
+    """Global shared access cache for provider data."""
+
+    _instance: Optional["GlobalCache"] = None
```
ZeldaZach, line 223: Can we not re-use the @singleton approach we've been using?

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+            self._card_kingdom = CardKingdomProvider()
+        return self._card_kingdom
+
+    @property
+    def cardmarket(self) -> CardMarketProvider:
+        if self._cardmarket is None:
+            self._cardmarket = CardMarketProvider()
+        return self._cardmarket
+
+    def load_all(self, force_refresh: bool = False) -> "GlobalCache":
```
ZeldaZach, line 296: We're mixing String annotations and normal. Pick one style

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+        """Load all provider data and pre-compute aggregations."""
+        LOGGER.info("Loading global cache data...")
+        
+        self._load_sets_metadata()
+
+        with ThreadPoolExecutor(max_workers=6) as executor:
```
ZeldaZach, line 302: Would abstracting the TPE make sense? Similar to what we did with the parallel_call op?

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+        self.meld_triplets = load_json("meld_triplets.json")
+        self.wc_signatures = load_json("world_championship_signatures.json")
+        self.mkm_translations = load_json("mkm_set_name_translations.json")
+        self.base_set_sizes = load_json("base_set_sizes.json")
+        LOGGER.info("Resource overrides loaded")
+
+    def _load_multiverse_bridge_cards(self) -> dict[str, list[dict[str, Any]]]:
+        """Load Multiverse Bridge Rosetta Stone cards."""
+        LOGGER.info("Loading Multiverse Bridge Rosetta Stone cards...")
+        from mtgjson5.providers import MultiverseBridgeProvider
```
ZeldaZach, line 361: I'm not a huge fan of doing imports outside of top of file, is there a benefit to this?

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+           
+        # Foreign data: (set, collector_number) -> list of foreign entries
+        self.foreign_data_map = (
+            extract_foreign_data(self.cards_df.lazy())
+            .collect()
+        )
+        
+    def _load_uuid_cache(self) -> None:
+        """Load legacy UUID mappings if available."""
+        cache_path = constants.RESOURCE_PATH / "legacy_mtgjson_v5_uuid_mapping.json"
```
ZeldaZach, line 418: I prefer joinpath

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+        if not cache_path.exists():
+            LOGGER.info("No UUID cache found")
+            return
+        try:
+            self.uuid_cache_df = pl.read_json(cache_path)
+            LOGGER.info(f"Loaded UUID cache with {len(self.uuid_cache_df)} entries")
+        except Exception as e:
```
ZeldaZach, line 425: Can we narrow down the Exception type?

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+        self.multiverse_bridge_map = (
+            MultiverseBridgeProvider().get_rosetta_stone_cards()
+        )
+        self.cardhoarder_map = CardMarketProvider().get_cardmarket_price_map()
+
+        self.uuid_cache_df = self._load_uuid_cache()
+
+    def _load_uuid_cache(self) -> pl.DataFrame | None:
+        """Load legacy UUID mappings if available."""
+        cache_path = Path("resources/legacy_mtgjson_v5_uuid_mapping.json")
```
ZeldaZach: There should be a constant for "resources" i think?

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+    def get_foreign_data(
+        self, set_code: str, collector_number: str
+    ) -> list[dict[str, Any]]:
+        """Get foreign language data for a specific card."""
+        if self.foreign_data_map is None:
+            return []
+
+        result = self.foreign_data_map.filter(
+            (pl.col("set") == set_code.upper())
+            & (pl.col("collector_number") == collector_number)
```
ZeldaZach, line 443: You're using a bitwise comparison? Can you add a comment on why?

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+        )
+
+        if result.is_empty():
+            return []
+
+        # Extract and force to Python list
+        foreign_data = result["foreign_data"].to_list()[0]
```
ZeldaZach, line 450: What if list size is 0? This could crash

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+            if triplet[2] == result_name:
+                return triplet[:2]
+        return None
+
+    @classmethod
+    def get_instance(cls) -> "GlobalCache":
+        """Get the singleton instance of GlobalCache."""
+        if cls._instance is None:
+            cls()
+        assert cls._instance is not None
```
ZeldaZach, line 469: asserts should not be in production code, only test code. You can throw an exception if there's a problem, but don't use assert

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,429 @@
...
+        return cls._instance
+
+
+global_cache = GlobalCache.get_instance()
```
ZeldaZach: Can we call this GLOBAL_CACHE instead? Also, how does this interact with the current cache scheme?

mtgjson5/providers/github/__init__.py
```diff
@@ -0,0 +1,68 @@
+from dataclasses import dataclass, field
```
ZeldaZach: We have a mix of from mtgjson5 and from ., we should stay consistent. Also, I prefer import X instead for certain stuff

mtgjson5/providers/github/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        provider_map: dict[str, GithubProviderType] = {
+            "decks": self.decks,
+            "sealed": self.sealed,
+            "boosters": self.boosters,
+            "mtgsqlite": self.mtgsqlite,
+            "card_sealed_products": self.card_sealed_products,
+        }
+        return provider_map[provider_name]
+    
+    def get_all_providers(self) -> list[GithubProviderType]:
```
ZeldaZach: I haven't gotten far enough, but why a list and not a dict?

mtgjson5/set_builder.py
```diff
@@ -45,90 +48,77 @@
...
+    
+    @property
+    def is_multi_face(self) -> bool:
+        return bool(self.scryfall.get("card_faces"))
+    
+    @property
+    def face_data(self) -> dict[str, Any]:
+        """Get the relevant face data, or the card itself for single-face."""
+        if self.is_multi_face:
+            return self.scryfall["card_faces"][self.face_id]
```
ZeldaZach: This could crash if index out of bounds

mtgjson5/set_builder.py
```diff
@@ -45,90 +48,77 @@
...
+    @property
+    def face_count(self) -> int:
+        if self.is_multi_face:
+            return len(self.scryfall["card_faces"])
+        return 1
+
 
 def parse_foreign(
-    sf_prints_url: str, card_name: str, card_number: str, set_name: str
+    set_code: str, card_number: str, _card_name: str
```
ZeldaZach, line 97: params shouldn't start with a _

mtgjson5/set_builder.py
```diff
@@ -298,21 +276,22 @@ def parse_legalities(sf_card_legalities: Dict[str, str]) -> MtgjsonLegalitiesObj
     return card_legalities
 
 
-def parse_rulings(rulings_url: str) -> List[MtgjsonRulingObject]:
+def parse_rulings(oracle_id: Optional[str]) -> List[MtgjsonRulingObject]:
```
ZeldaZach, line 294: str | None over Optional (I'll probably do a wide sweep of this in the near future)

mtgjson5/set_builder.py
```diff
@@ -640,75 +635,71 @@ def build_base_mtgjson_cards(
 
 
 def add_is_starter_option(
-    set_code: str, search_url: str, mtgjson_cards: List[MtgjsonCardObject]
+    set_code: str, _search_url: str, mtgjson_cards: List[MtgjsonCardObject]
```
ZeldaZach, line 653: Don't use _ in param names

mtgjson5/set_builder.py
```diff
@@ -640,75 +635,71 @@ def build_base_mtgjson_cards(
...
-    starter_cards = ScryfallProvider().download(starter_card_url)
 
-    if starter_cards["object"] == "error":
+    if global_cache.cards_df is None:
+        LOGGER.warning(f"Cache not loaded, skipping starter data for {set_code}")
+        return
+
+    # Get scryfall IDs for non-booster cards in this set
+    starter_ids = set(
+        global_cache.cards_df.filter((pl.col("set") == set_code.upper()) & ~pl.col("booster"))[
```
ZeldaZach: You're doing some wonky stuff with bitwise ops, can we get more context/comment on why?

mtgjson5/set_builder.py
```diff
@@ -798,484 +789,511 @@ def add_extra_language_uuids(
...
+    card.identifiers.mtgo_foil_id = get_str_or_none(sf.get("mtgo_foil_id"))
+    
+    # Multiverse ID
+    multiverse_ids = sf.get("multiverse_ids") or []
+    if multiverse_ids:
+        idx = ctx.face_id if ctx.face_id < len(multiverse_ids) else 0
+        card.identifiers.multiverse_id = get_str_or_none(multiverse_ids[idx])
+
+
+def _set_face_data(card: MtgjsonCardObject, ctx: CardBuildContext) -> None:
```
ZeldaZach, line 892: ctx is a weird variable name

mtgjson5/set_builder.py
```diff
@@ -1396,8 +1414,15 @@ def add_card_kingdom_details(mtgjson_set: MtgjsonSetObject) -> None:
     :param mtgjson_set: MTGJSON Set
     """
     LOGGER.info(f"Adding CK details for {mtgjson_set.code}")
-    translation_table = CardKingdomProvider().get_scryfall_translation_table()
-
+    if global_cache.card_kingdom_map is None:
+        LOGGER.warning("Card Kingdom cache not loaded")
```
ZeldaZach, line 1412: Might want to differentiate the warnings as it's not clear if the cache was not set or the ttable was not set

mtgjson5/set_builder.py
```diff
@@ -1622,20 +1652,22 @@ def get_base_and_total_set_sizes(mtgjson_set: MtgjsonSetObject) -> Tuple[int, in
...
-                base_set_size_download = ScryfallProvider().download(
-                    ScryfallProvider().CARDS_IN_SET.format(mtgjson_set.code.upper())
-                )
-
-            base_set_size = int(base_set_size_download.get("total_cards", 0))
+            # Count cards in base set from cache
+            if global_cache.cards_df is not None:
+                base_set_count = global_cache.cards_df.filter(
+                    (pl.col("set") == mtgjson_set.code.upper())
+                    & (pl.col("lang") == "en")
```
ZeldaZach, line 1653: Can we get more context on the bitwise ops?

mtgjson5/set_builder.py
mtg-elmo: No im not

mtgjson5/set_builder.py
mtg-elmo: i dont see any...?

the-muppet: Polars (and Pandas) use & for element-wise logical AND between boolean Series/Expressions because Python's and keyword can't be overloaded.

mtgjson5/cache_builder.py
the-muppet: would you prefer I import __annotations__ or use Self from typing?

ZeldaZach: I'm good with annotations

mtgjson5/providers/scryfall/monolith.py
the-muppet: That could damage or shatter the identity of your monolith file, for what is a monolith if stripped of the things that compose it?

...sadness, zach. the monolith is sadness.

ZeldaZach: The monolith needs to die, and this is one way to get rid of it

mtgjson5/set_builder.py
the-muppet: shame on you for dunking on Golang in public

mtgjson5/cache_builder.py
the-muppet: Hmm.. what do you mean?

mtgjson5/cache_builder.py
the-muppet: I mean. *we can*, the singleton_decorator obfuscates underlaying class objects and typing though.. no IDE support, etc. I could add a Singleton metaclass to the ABC instead?

mtgjson5/providers/scryfall/monolith.py
the-muppet: There are many different ways to do the thing in Python, its so expressive! We sure are lucky, huh?

mtgjson5/set_builder.py
```diff
@@ -798,482 +804,488 @@ def add_extra_language_uuids(
...
+    if not face_names:
+        return
+
+    card.face_name = str(face.get("name", ""))
+
+    if card.layout == "meld":
+        return  # Meld handled separately
+
+    # Special case: TUST tokens
+    if card.set_code.upper() == "tust":
```
ZeldaZach, line 980: did you mean .lower()? This will never match currently

mtgjson5/set_builder.py
```diff
@@ -798,482 +804,488 @@ def add_extra_language_uuids(
...
-    )
-    mtgjson_card.printings = parse_printings(
-        scryfall_object["prints_search_uri"].replace("%22", "")
-    )
-    mtgjson_card.legalities = parse_legalities(
-        scryfall_object["legalities"]
-        if scryfall_object.get("set_type") not in ["memorabilia"]
-        else {}
+    # ADSK special case
+    if card.set_code.upper() == "adsk":
```
ZeldaZach, line 1005: again should be lower

mtgjson5/providers/scryfall/monolith.py
```diff
@@ -11,6 +11,7 @@
 from typing import Any, Dict, List, Optional, Set, Union
```
ZeldaZach, line 11: We can revert this file, since this is all just spacing changes and an unused import

mtgjson5/cache_builder.py
```diff
@@ -0,0 +1,473 @@
+"""
+Global cache for MTGJSON provider data and pre-computed aggregations.
+"""
+
+import json
+import ijson
```
ZeldaZach, line 6: Do we need both json and ijson? If so, you'll need to add ijson to the requirements

PR #1350, opened 2025-11-27
Adds flake8 import style plugin for MTGJSON

the-muppet:
Introduces a custom flake8 plugin to enforce consistent import
style patterns across the codebase, allowing only whitelisted modules for 'from ... import ...' statements.
Updates configuration files to integrate plugin with existing linting workflow for improved code quality and cause honestly, i'm never gonna remember

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,29 @@
+[tool.black]
+line-length = 88
+target-version = ["py39",
+"py310",
+"py311",
+"py312",
+"py313"]
+
+[tool.isort]
+profile = "black"
+
+
+# 501 - line too long
+# E402 - module level import not at top of file
+# W503 - line break before binary operator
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+# E302 - expected 2 blank lines, found N
+# W291 - trailing whitespace
+[tool.flake8]
+extend-ignore = [
+    "E501",
+    "E402",
+    "W503",
+    "W293",
+    "E302",
+    "W291"
```
ZeldaZach, line 26: Any reason we picked these to ignore? 501 is fine. 402 I'm iffy on, but we can make it work. 503 is fine to skip. 293 i'd like to enforce. 302 i'd like to enforce. 291 i'd like to enforce.

PR #1364, opened 2026-01-11
fix(decks): use deck's
    set_code for sealedProductUuids lookup

the-muppet:
## Summary
   - Fixed bug where  lookup used the function parameter  instead of each deck's actual 
   - This caused only the first requested set's decks to get their sealedProductUuids populated


the-muppet:
I hate tox. this was just a formality, really.. but.. gah. (╯°□°）╯︵ ┻━┻

ZeldaZach (approved):

PR #1365, opened 2026-01-12
fix(promoTypes): prevent duplicate planeswalkerstamped for multi-face cards

the-muppet:
## Summary
  - Multi-face cards (adventures, split cards, etc.) were getting `planeswalkerstamped` appended once per face when collector number ends in 'p'
  - This resulted in duplicates like `['promopack', 'stamped', 'planeswalkerstamped', 'planeswalkerstamped']`
  - Fix: Check if already present before appending

  ## Affected Cards
  47 cards across PAKH, PDSK, PWOE, PELD, PLCI, PMKM, PGRN, PMOM sets.


ZeldaZach:
Can you re-base this PR? Looks like it has a lot of unrelated changes

PR #1368, opened 2026-01-28
Add produced_mana field

Emnalyeriar:
Hey, I added the `produced_mana` field to MTGJson cards, which is filled from Scryfall data. It uses face data first when a card has multiple faces.
Should I add any tests for this? Can I test this locally?
I am not sure if there is a way for me run the project locally.

### DIFF
--- a/mtgjson5/classes/mtgjson_card.py
+++ b/mtgjson5/classes/mtgjson_card.py
@@ -34,6 +34,7 @@ class MtgjsonCardObject(JsonObject):
     color_identity: List[str]
     color_indicator: Optional[List[str]]
     colors: List[str]
+    produced_mana: Optional[List[str]]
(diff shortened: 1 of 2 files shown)


ZeldaZach (approved):

Emnalyeriar:
@ZeldaZach One thing I thought of now is that the mdfc lands, where the front side is a spell, and the back side is a land, will have the produced_mana field on both of the sides. 
https://scryfall.com/card/znr/90/agadeems-awakening-agadeem-the-undercrypt
Scryfall provides the produced_mana field on the main card object, not on specific card faces. In mtgjson, each face (side) is a separate entry, and both of those entries will have produced_mana field (if I understand how this works correctly). Not sure if this is an issue or not?

PR #1447, opened 2026-02-06
Add card.tokens which link to the UUIDs within a set that the token creates

### LINKED ISSUE #1189: Cards do not have the relatedCards.reverseRelated fields for any tokens
#### Version tested against
MTGJSON v5.2.2+20240310

#### Description of Bug
The documentation tells that cards should have a relatedCards.reverseRelated field for any token created by the card. However, this is not set in the data for any card.

ZeldaZach:
- Fix #1189

Example: [EOE.json](https://github.com/user-attachments/files/25120566/EOE.json)

### DIFF
--- a/mtgjson5/v2/pipeline/core.py
+++ b/mtgjson5/v2/pipeline/core.py
@@ -1660,23 +1660,16 @@ def add_reverse_related(lf: pl.LazyFrame) -> pl.LazyFrame:
 
 
 # 4.4: add token UUIDs to non-token cards
-def add_token_ids(lf: pl.LazyFrame, scryfall_to_uuids: dict[str, list[str]]) -> pl.LazyFrame:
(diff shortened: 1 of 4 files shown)


mtgjson5/v2/pipeline/core.py
```diff
@@ -1656,7 +1656,43 @@ def add_reverse_related(lf: pl.LazyFrame) -> pl.LazyFrame:
...
+            pl.when(pl.element().struct.field("component") == "token")
+            .then(pl.element().struct.field("id"))
+            .otherwise(pl.lit(None))
+        )
+        .list.drop_nulls()
+        .alias("_token_scryfall_ids")
+    )
+
+    # Resolve Scryfall IDs to MTGJSON UUIDs via dict lookup
+    return lf.with_columns(
```
the-muppet2: using a explode-join-aggregate method is the 'correct' approach here. you can derive scryfall_uuid mapping once from cards_lf (cached on ctx instead of per card) to do this in a single pass

PR #1506, opened 2026-02-17
reorganize modules to live directly under mtgjson5

the-muppet2:


mtgjson5/v2/_version.py
```diff
@@ -1,8 +0,0 @@
-"""Dynamic version read from mtgjson.properties."""
-
-import configparser
```
ZeldaZach, line 3: Why'd we get rid of this?

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mtgjson5"
-dynamic = ["version"]
```
ZeldaZach, line 7: intentional?

PR #1556, opened 2026-03-03
Memory and performance optimization for the MTGJSON build pipeline

the-muppet2:
Subprocess isolation - Polars uses jemalloc which never returns freed memory to the OS. All heavy phases now run in isolated subprocesses, so allocations are fully reclaimed on process exit:
  - Assembly tasks split into 6 groups (A–F) running as subprocesses with max 2 concurrent, reducing peak RSS
  - Exports and prices run in their own subprocess, with prices nested in parallel alongside format exports (sqlite, csv, sql, psql)
  
Pipeline batching - build_cards() processes sets in configurable batches with strategic frame releases between batches, preventing unbounded memory growth during the card pipeline.

Compression parallelization - ProcessPoolExecutor with per-format task granularity for large files. All 4 cores stay busy instead of 3 cores idling while one finishes xz compression on a large file. Tasks sorted by estimated CPU cost to prevent tail starvation.

Caching - CardMarket finishes and TCGPlayer SKUs cached to parquet during the pipeline, so export subprocesses skip re-computation.

Price build dedup - Fixed double provider fetch when both parquet and JSON prices are built. today_df from the parquet price build is now passed to the JSON price build, eliminating redundant network calls.

Profiling - New --profile flag with PipelineProfiler and SubprocessProfiler classes. Records wall-clock timing and RSS at named checkpoints across parent and all child processes. Outputs profile_report.json and profile_summary.log.

QoL - --build-all now includes --compress and --price-build. Log files consolidated across subprocesses into a single log per build.


mtgjson5/build/formats/parquet.py
```diff
@@ -14,7 +14,7 @@
     from ..context import AssemblyContext
 
 _COMPRESSION: Literal["zstd"] = "zstd"
-_COMPRESSION_LEVEL = 9
+_COMPRESSION_LEVEL = 3
```
ZeldaZach, line 17: is this a significant change?

ZeldaZach (approved):

PR #1637, opened 2026-04-10
feat: add CardTrader provider

Emnalyeriar:
This PR adds CardTrader to the v2 price build.

I tried to follow the existing provider patterns and based the implementation on the other price providers already in the repo. I also added error handling, so CardTrader failures should not fail the whole price build. I do not fully understand the repo rules, and as far as I know I'm not able to run the full build locally. Being honest, I used AI to help me understand the codebase and figure out the implementation.

For the CardTrader market price, this implementation calculates the price as the average of the first 15 returned marketplace listings. As far as I understand, this matches how CardTrader calculates its own market price, but I think this needs to be discussed and tested.

What I tested locally:
- PYTHONPATH=. uv run --extra dev pytest -q tests/mtgjson5/providers/test_cardtrader.py
- PYTHONPATH=. uv run --extra dev python -m mtgjson5 -s MH3 --skip-mcm --no-alerts
- PYTHONPATH=. uv run --extra dev python -m mtgjson5 --price-build --skip-mcm --no-alerts
- PYTHONPATH=. uv run --extra dev tox -e format-check,lint,mypy,unit

If there is a better way to validate the changes I did, please let me know :)


the-muppet2:
@ZeldaZach - wanted to get your take on whether we want to bring in the CardTrader provider, and if so, how to integrate it. I pulled the branch and ran the provider in isolation:

## Results

| Metric | Value |
|--------|-------|
| Total fetch time | **~19 minutes** |
| Rows returned | 188,682 (94,341 normal + 94,341 foil) |
| Unique scryfallIds | 94,165 |
| Currency | 100% USD |
| API calls | ~2,274 (3 per expansion) |
| Price range | $0.14 - $10,000,000.70 |
| Median price | $0.41 |

I assume the $10M max is a junk listing - the averaging approach (first 15 listings per blueprint) doesn't filter outliers. Median of $0.41 looks reasonable for a marketplace-wide average.

### Code Quality

The implementation is well-structured overall - config pattern matches our existing providers, test coverage is solid (7 tests), and the Phase 1/Phase 2 raw-parquet-to-UUID-mapping plumbing is already wired into `price_fetcher.py` and `price_builder.py` correctly.

Issues worth addressing before merge:

- No retry logic on transient failures - a single 429 or timeout silently drops an entire expansion's price data via `raise_for_status()` in the outer try/except
- No `generate_today_price_dict()` - breaks the established provider interface that every other price provider implements
---
- No checkpoint/resume - if the ~19-minute fetch is interrupted, all progress is lost and restarts from scratch. Our other API heavy providers (MCM & TCG) have checkpointing in place 
- No `User-Agent` header - other providers identify as `MTGJSON/5.0`
- Currency validation was rejecting records as ERROR then silently continuing - records should (probably?) flow through with the currency column preserved
---
- Float equality in test should use `pytest.approx`
- Missing `fetch_prices_sync` wrapper (all other providers have one)

### Pipeline Integration Concern

At ~19-20 minutes, this is potentially too slow for inline pipeline execution. The background fetch does run concurrently with the card build, but `__main__` would block for ~15+ minutes waiting on CardTrader to finish.

### Proposal: CardMarket-style S3 Pattern

CardMarket has the same slow fetch problem and solves it by decoupling the fetch from the pipeline:

1. Run the raw price fetch as a **separate cron job**
2. Store `cardtrader_raw_prices.parquet` in **S3** alongside the CardMarket raw data
3. Pipeline reads the cached parquet from S3 at build time - Phase 2 mapping (raw to UUID join) takes milliseconds
4. Add checkpoint/resume so partial runs aren't wasted

The Phase 2 plumbing in `price_builder.py` is already implemented in this PR, so the mapping side just works. Only Phase 1 needs consideration for moving out of the hot path.

### Questions
1. Do we want CardTrader pricing data in MTGJSON at all? At ~94K unique cards the coverage is good, and it's another paper retail price source.
2. If yes - are we okay adding ~19 minutes to the build, or should we require the CardMarket-style S3 pattern before merging?
3. Also if yes - should we filter/flag potential outliers? It's not really "our job" to scope that, but still worth considering
4. Should we ask @Emnalyeriar to add checkpoint/resume and retry logic, or handle that ourselves as a follow-up?

Emnalyeriar:
Thank you for reviewing this! I think the outliers need to be addressed definitely. I can take a look and address some of the concerns, not all of them unfortunately, given I cannot run this locally.
While reading this I figured out that I didn't handle identifiers and purchase urls, should those be handles as well in this PR?

Emnalyeriar:
@the-muppet2 Why is the currency in your summary USD? The expected currency should be EUR

the-muppet2:
I assume because my CardTrader account is USD - Every blueprint was failing with 'Expected EUR got USD'

the-muppet2:
> Thank you for reviewing this! I think the outliers need to be addressed definitely. I can take a look and address some of the concerns, not all of them unfortunately, given I cannot run this locally. While reading this I figured out that I didn't handle identifiers and purchase urls, should those be handles as well in this PR?

Well - Identifiers are _partially_ handled. The blueprint_id --> scryfall_id mapping is built during the fetch (_build_blueprint_map), and cardtraderId is already defined in the schema in submodels.py. But the blueprint_id is used and discarded, it's only used as a join key to look up prices, then thrown away. It never gets persisted to the raw output or wired into the card's identifiers struct.

So, the data is there during the fetch, it's just not being kept. To actually get cardtraderId onto cards would need:
1. Include blueprint_id in the raw records output - it's right there, just not appended
2. Wire it into add_identifiers_struct() in identifiers.py
3. Fix the schema classification - it's currently listed under sealed product identifiers in submodels.py instead of card identifiers

Purchase URLs are a separate concern entirely - not within scope of this PR

Emnalyeriar:
> > Thank you for reviewing this! I think the outliers need to be addressed definitely. I can take a look and address some of the concerns, not all of them unfortunately, given I cannot run this locally. While reading this I figured out that I didn't handle identifiers and purchase urls, should those be handles as well in this PR?
> 
> Well - Identifiers are _partially_ handled. The blueprint_id --> scryfall_id mapping is built during the fetch (_build_blueprint_map), and cardtraderId is already defined in the schema in submodels.py. But the blueprint_id is used and discarded, it's only used as a join key to look up prices, then thrown away. It never gets persisted to the raw output or wired into the card's identifiers struct.
> 
> So, the data is there during the fetch, it's just not being kept. To actually get cardtraderId onto cards would need:
> 
> 1. Include blueprint_id in the raw records output - it's right there, just not appended
> 2. Wire it into add_identifiers_struct() in identifiers.py
> 3. Fix the schema classification - it's currently listed under sealed product identifiers in submodels.py instead of card identifiers
> 
> Purchase URLs are a separate concern entirely - not within scope of this PR

I poked the code a little bit more with the help of AI, I tried to persist the blueprint_id as cardtraderId. Afaik, there is an additional fetch required `fetch_identifier_lookup` but maybe this can be avoided somehow?

Emnalyeriar:
> ## Results
> Metric	Value
> Total fetch time	**~19 minutes**
> Rows returned	188,682 (94,341 normal + 94,341 foil)
> Unique scryfallIds	94,165
> Currency	100% USD
> API calls	~2,274 (3 per expansion)
> Price range	$0.14 - $10,000,000.70
> Median price	$0.41
> I assume the $10M max is a junk listing - the averaging approach (first 15 listings per blueprint) doesn't filter outliers. Median of $0.41 looks reasonable for a marketplace-wide average.
> 

@the-muppet2, can you tell me which cards had the price of $10M? I want to research how to properly handle the outliers and having an example would be helpful

the-muppet2:
My tests were run from a temp directory, so I don't have that parquet anymore

Emnalyeriar:
> My tests were run from a temp directory, so I don't have that parquet anymore

Is there any way I can run this locally to inspect the result?

Emnalyeriar:
Update from my side @the-muppet2 

I reworked the CardTrader pricing fetch/calculation based on what we discussed

Changes made:
- Fixed foil pricing: normal and foil prices are now calculated separately using CardTrader’s actual foil flag from the marketplace listings
- Updated market price calculation so obvious outliers are filtered before averaging
- Added basic safeguards for unreliable price samples, including ignoring very thin listing groups and listings that should not represent normal market price
- Kept the calculation based on CardTrader marketplace listings, but made the resulting prices more stable and representative

I also generated and inspected a fresh raw parquet locally. The previous issue where normal/foil prices were duplicated is fixed, and the extreme multi-million outliers are gone.

I think the current filtering is reasonable, but I am open to a follow-up discussion :)

the-muppet2:
@ZeldaZach

re-ran the isolation test on the same setup as before. Most of the issues from my initial review have been addressed. There are a couple of things worth discussing though.

### Results comparison

| Metric | Original | Updated PR | Result |
|---|---|---|---|
| Fetch time | ~19 min | **~50 min** | 2.6× slower |
| Total records | 188,682 | 130,947 | −31% |
| Unique scryfallIds | 94,165 | 85,578 | −9% |
| Max price | $10,000,000.70 | $46,512.54 | much better |
| Median | $0.41 | $0.49 | ≈ |
| WARNINGs | (low) | 9 (all genuine high-value cards) | ok |
| ERRORs | 0 (functionally; all data was junk) | **10,382 GBP currency mismatches** | new issue |

### Blockers requiring discussion

**1. Runtime regressed 2.6×, despite dropping request delay from 1.05s -> 0.13s.**

The cause is the new fallback in `_fetch_blueprint_finish_listings`: when an expansion-level finish group has <3 eligible listings or a >10× median outlier, the provider fires an extra per-blueprint call, there were roughly **~22K additional fallback calls** in my run.

The structural concern I raised in my first review (~19 min inline + no resume) is now **strictly worse**. A few options:

- Cap the fallback rate per build, i.e. skip targeted retries once a per-expansion threshold is hit
- Move to the CardMarket-style pattern: raw fetch as a separate cron, parquet cached to S3, pipeline reads it
- Add checkpoint/resume so a 50-min run isn't a total loss on crash (more nice to have than anything tbh, option 2 is likely how this has to be added)

**2. Currency mismatch silently drops ~10K records (all GBP in my run).**

`_build_raw_record` still does `LOGGER.error(...)` then `return None` on any blueprint whose listing slice returns a non-`expected_currency`. Better behavior would be to keep the record and preserve `currency` on the row. the schema already supports it, and downstream consumers can normalize. As-is, ~8% of unique cards are lost on a USD account just because some sellers price in GBP.

ZeldaZach:
I appreciate the breakdown @the-muppet2 it's very insightful and helps drive direction for me.

With this PR, if it takes almost an hour to get the prices from one provider on a daily basis, that's pretty significant. Is there anyway we can reduce down the time? I'm not sure if CT has any kind of bulk-style endpoints we can maybe take advantage of for these operations, do things in parallel, etc. If we can get this down to ~10 mins, I'd consider it fully.

Not sure why total records dropped so much, maybe we need to account for all currencies and inject them into the records at a lower level than we do now? Might require some re-architecture which I'm not thrilled about.

If you need help furthering this, do let us know!

Emnalyeriar:
Thank you for reviewing and checking my changes :) I’m happy to work on this further, and as soon as I have some more time, I'll improve the CardTrader provider based on the suggestions!

Emnalyeriar:
@the-muppet2 @ZeldaZach I dug into the CardTrader API behavior more because the runtime regression is tied to finish-specific fallback calls.

Fetching marketplace data by expansion returns the cheapest 25 products per blueprint. A blueprint is CardTrader’s card-printing/product definition, roughly the CardTrader-side identifier for a specific Scryfall card/printing. Because the expansion response is capped to the cheapest products per blueprint, it can under-cover foil prices when nonfoil listings occupy those cheapest 25 products.

So I think there are two viable paths:

1. Keep CardTrader inline and fast:
Fetch once per expansion, group by `properties_hash.mtg_foil`, calculate prices only when the returned finish group has enough eligible listings, and skip thin/missing groups. This is conservative and avoids normal/foil duplication, but it does not maximize foil coverage.

2. Maximize CardTrader coverage:
Keep the fallback logic and call the blueprint endpoint with `blueprint_id + foil=true/false` for missing/thin finish groups, but move the raw fetch out of the hot build path with a CardMarket-style S3/cache/checkpoint flow.

What do you think?

PR #1640, opened 2026-04-15
feat: add source field to deck models

Emnalyeriar:

### DIFF
--- a/mtgjson5/models/decks.py
+++ b/mtgjson5/models/decks.py
@@ -62,7 +62,7 @@ class DeckList(PolarsMixin, BaseModel):
     )
     source: str | None = Field(
         default=None,
-        description="A source URL for the deck list.",
+        description="A human-readable source for the deck list.",
         json_schema_extra={"optional": True},
     )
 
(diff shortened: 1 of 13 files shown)


mtgjson5/models/decks.py
```diff
@@ -60,6 +60,11 @@ class DeckList(PolarsMixin, BaseModel):
         ),
         json_schema_extra={"introduced": "v4.3.0"},
     )
+    source: str | None = Field(
+        default=None,
+        description="A source URL for the deck list.",
```
taw: It's not necessarily URL. It should be treated as human-readable only.

It can be one of:
 
* real URL that actually resolves
* URL that no longer works (generally because WotC nukes their website regularly), so you can get it through web archive
* some comment explaining where data comes from

Some random examples of non-URL sources: 

```
data/arena-starter/otj/Second Chance.txt:// SOURCE: MTG Arena
data/redemption/ogw/Oath of the Gatewatch Redemption.txt:// SOURCE: database query with -number:/a/
data/redemption/ons/Onslaught Foil Redemption.txt:// SOURCE: default database query
data/shandalar/past/Witch.txt:// SOURCE: 0056.DCK from game files
data/theme/ps11/Eons of Evil.txt:// SOURCE: email with photos from Santiago Guerra
```

It's also often not complete, like printing information often comes from deck opening videos, but I didn't normally add second URL for it (sometimes it's in git commit message). If people care much about it, we could add multiple URLs. Or I could have fixed obvious typos etc. based on best judgement.

Emnalyeriar: Thanks, updated the description to `A human-readable source for the deck list.`

PR #1648, opened 2026-04-30
Support new CardSphere Approach

ZeldaZach:
CardSphere recently injected SFIDs into their outputs, which makes removing (most of) multiversebridge possible. This is my approach for CardSphere integration directly via their site.


mtgjson5/providers/cardsphere/provider.py
```diff
@@ -0,0 +1,274 @@
...
+    ) -> list[dict[str, str | None]]:
+        """Download and parse a single set's CSV.
+
+        Returns list of dicts with keys: scryfallId, cardsphereId, foil.
+        """
+        url = CSV_URL_TEMPLATE.format(set_id=set_id)
+        response = session.get(url, timeout=30)
+        response.raise_for_status()
+
+        rows: list[dict[str, str | None]] = []
```
the-muppet2: Hot-path Python loop - please move CSV parsing to polars.

This is per-row Python work that drags pipeline performance. csv.DictReader + .strip() + dict construction runs O(rows × csvs) in Python, and the result feeds pl.DataFrame(all_rows), which is the slowest possible way to build a polars frame.

instead:
return a pl.DataFrame instead of a list[dict]:
```python
  df = (
      pl.read_csv(
          io.BytesIO(response.content),
          columns=["Scryfall ID", "Cardsphere ID", "Foil"],
      )
      .rename({"Scryfall ID": "scryfallId", "Cardsphere ID": "cardsphereId", "Foil": "foil"})
      .filter(pl.col("scryfallId").str.len_bytes() > 0)
      .filter(pl.col("cardsphereId").str.len_bytes() > 0)
  )
  return df
```
Then in fetch_and_build, collect a list[pl.DataFrame] and pl.concat(frames) instead of pl.DataFrame(all_rows). CSV parsing now runs in Rust, no per-row dict allocation, no schema inference cost.

mtgjson5/providers/cardsphere/provider.py
```diff
@@ -0,0 +1,274 @@
...
+                    "cardsphereSetName": s["name"],
+                }
+                for s in sets_list
+            ],
+            schema=SET_SCHEMA,
+        )
+
+        # Download all CSVs concurrently and collect rows
+        all_rows: list[dict[str, str | None]] = []
+        session = _build_session()
```
the-muppet2: Session leaks if anything between 101-118 raises. requests.Session is a context manager - wrap it:
```python
with _build_session() as session:
```
and drop session.close()

mtgjson5/providers/cardsphere/provider.py
```diff
@@ -0,0 +1,274 @@
...
+
+        set_ids = [s["id"] for s in sets_list]
+        failed = 0
+
+        with ThreadPoolExecutor(max_workers=_MAX_WORKERS) as executor:
```
the-muppet2: how long does this take? this kicks off with 9 other siblings and `as_completed` in cache.py waits for the slowest one. 

If this is a long running operation, decoupling cardsphere latency from bulk-load path can be achieved by kicking it off in background and only block in consolidate_lookups when we actually need it, like _start_tcg_skus_fetch / _await_tcg_skus operates

mtgjson5/providers/cardsphere/provider.py
```diff
@@ -0,0 +1,274 @@
...
+            futures = {executor.submit(self._download_csv, session, set_id): set_id for set_id in set_ids}
+
+            for future in as_completed(futures):
+                set_id = futures[future]
+                try:
+                    rows = future.result()
+                    all_rows.extend(rows)
+                except Exception as e:
```
the-muppet2: Silent at DEBUG, only an aggregate WARNING on 121.. If 50/800 sets fail, we'll never know which without flipping log levels. At minimum, keep a small list of failed set_ids and log them once at WARNING so triage is possible.

mtgjson5/providers/cardsphere/provider.py
```diff
@@ -0,0 +1,274 @@
...
+            LOGGER.warning(f"Failed to download {failed}/{len(set_ids)} CSVs")
+
+        LOGGER.info(f"Downloaded {len(all_rows):,} card rows from CardSphere")
+
+        if not all_rows:
+            self._cards_df = self._empty_cards_df()
+            return self._cards_df, self._sets_df
+
+        # Build raw DataFrame from all rows
+        raw_df = pl.DataFrame(all_rows)
```
the-muppet2: Pass an explicit schema so this doesn't break on degenerate inputs.
Becomes moot if you adopt the pl.read_csv + pl.concat refactor above.

mtgjson5/providers/cardsphere/provider.py
```diff
@@ -0,0 +1,274 @@
...
+            response = requests.get(SETS_URL, timeout=30, headers={"User-Agent": "MTGJSON/5.0"})
+            response.raise_for_status()
+        except Exception as e:
+            LOGGER.error(f"Failed to fetch CardSphere sets page: {e}")
+            return []
+
+        html = response.text
+
+        # Look for __NEXT_DATA__ JSON payload
+        match = re.search(
```
the-muppet2: three regex fallbacks for the same data is fragile, and would all likely break together if CS tweaks HTML.

use BS4 or html.parser to find the script tag, then json.load its contents - fail loudly if not present

mtgjson5/providers/cardsphere/provider.py
```diff
@@ -0,0 +1,274 @@
...
+            )
+            .unique(subset=["scryfallId"], keep="first")
+        )
+
+        # Full outer join to capture cards that only exist in one finish
+        result = non_foil.join(foil, on="scryfallId", how="full", coalesce=True)
+
+        return result.select(
+            pl.col("scryfallId"),
+            pl.col("cardsphereId").cast(pl.String),
```
the-muppet2: Both columns are already pl.String from the join - these casts are no-ops. Can drop.

mtgjson5/data/cache.py
```diff
@@ -1190,6 +1199,30 @@ def _load_secretlair_subsets(self) -> None:
             sld_df.write_parquet(cache_path)
             self.sld_subsets_lf = sld_df.lazy()
 
+    def _load_cardsphere(self) -> None:
+        """Load CardSphere card and set ID mappings."""
+        cards_cache = self.cache_path / "cardsphere_cards.parquet"
```
the-muppet2, line 1200: filename mismatch:
- _load_cardsphere reads/writes cardsphere_cards.parquet (lines 1204, 1216)
- _dump_and_reload_as_lazy re-dumps under cardsphere.parquet (line 477)

mtgjson5/data/cache.py
```diff
@@ -1190,6 +1199,30 @@ def _load_secretlair_subsets(self) -> None:
...
+
+        if _cache_fresh(cards_cache) and _cache_fresh(sets_cache):
+            self.cardsphere_lf = pl.scan_parquet(cards_cache)
+            self.cardsphere_sets_lf = pl.scan_parquet(sets_cache)
+            return
+
+        try:
+            cards_df, sets_df = self.cardsphere.fetch_and_build()
+
+            if cards_df is not None and len(cards_df) > 0:
```
the-muppet2: This writes the parquets here, then `_dump_and_reload_as_lazy` writes them again at the end of `load_all`

the-muppet2:
No tests added. The network paths are hard, but `_pivot_to_card_mapping` and the `__NEXT_DATA__` parser in `_fetch_sets_index` are pure functions that take strings/frames and return frames; both worth a fixture-based test, especially since the latter is the brittle part of the integration

docs/pipeline-context.md
```diff
@@ -186,6 +187,8 @@ identifiers_lf = pl.LazyFrame({
     "cardKingdomId": str | None,
     "cardKingdomEtchedId": str | None,
     "cardKingdomFoilId": str | None,
+    "cardsphereId": str | None,
+    "cardsphereFoilId": str | None,
```
kodawah, line 192: no cardsphereEtchedId? do they manage etched cards differently? example STA or MH2 cards

kodawah:
+1 for tests!

mtgjson5/providers/cardsphere/provider.py
ZeldaZach: My tests of this have been about 2-3 minutes, but if there's a way to knock it to the background that could make sense

kodawah (approved):

kodawah:
1. you should wait for review after adding new commits
2. the approach is wrong, etched is not a finish

PR #1658, opened 2026-05-23
Add PDH and PreDH leadershipSkills

gldnbrwncouches:
These two formats use leaders/commanders in their game, but aren't yet present in the `leadershipSkills` struct.  This adds both of them with checks regarding their respective formats.

PDH (Pauper Commander) checks can all be done on basic values already found on the card similar to current Brawl, Oathbreaker, and Commander checks.  Scryfall legalities data didn't seem to include anything on legality "as a commander", and even though Scryfall does offer an `is:pauperCommander` search option, I didn't see that data anywhere.
- Example card as of this PR: ["Ajani's Pridemate"](https://scryfall.com/card/fdn/135/ajanis-pridemate) isn't legal in Pauper, doesn't show up in a Scryfall search for `legal:pauperCommander`, but is legal as the Commander if you search for it using `is:pauperCommander`.  [PDH rules here](https://pdhhomebase.com/rules/)

The best way I found to check for PreDH was to just check in `legalities`, which does require that stage to happen before the `leadershipSkills` in `relationships.py` does.  Luckily, that stage already does, I just wanted to point out the dependency.  Otherwise, it would need to be changed to check if a card's original printing was before 2011-06-17 (first commander product release date), and I didn't see a clear way to determine that at the current stage where `leadershipSkills` are done.
- Example card as of this PR: ["Jhoira of the Ghitu"](https://scryfall.com/card/mma/177/jhoira-of-the-ghitu) was printed in MMA (Modern Masters), and many other times after the cutoff date, but the original printing was in FUT (Future Sight) which is before the cutoff date, and therefore legal in the format and as a commander.

### DIFF
--- a/mtgjson5/models/submodels.py
+++ b/mtgjson5/models/submodels.py
@@ -369,12 +369,17 @@ class LeadershipSkills(TypedDict):
             "description": "If the card can be your commander in the [Pauper Commander](https://mtg.wiki/page/Pauper_Commander) format.",
             "introduced": "v5.3.1",
         },
+        "predh": {
+            "description": "If the card can be your commander in the [PreDH](https://articles.starcitygames.com/magic-the-gathering/commander-sub-format-predh-is-the-new-magic-rage/) format.",
+            "introduced": "v5.3.1",
+        },
     }
 
     brawl: bool
     commander: bool
     oathbreaker: bool
     pauper_commander: bool
+    predh: bool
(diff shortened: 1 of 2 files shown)


ZeldaZach (approved):

ZeldaZach:
Thanks for the contribution! Greatly appreciated :)

PR #1678, opened 2026-07-06
Fix missing non-foil TCGplayer SKUs for Surge Foil-anchored cards (MSC)

ZeldaZach:
> Supersedes #1677 (closed automatically when its slash-containing branch was renamed to `fix-msc-nonfoil-tcgplayer-skus` so the Docker container build gets a valid image tag). Same commits.

## Problem

Fixes #1676. For 52 MSC (Marvel Super Heroes Commander) cards, `TcgplayerSkus.json` only contained the Surge Foil product's foil SKUs — the Regular (non-foil) product and its SKUs were absent, even though the card's `finishes` array is `[nonfoil, foil]`.

Example — Tombstone, Career Criminal (0160), `uuid 474309ce-4270-5bc9-8078-f51d4b6ab395`:
- Present: productId `698031` (Surge Foil) + its foil skuIds
- Missing: productId `698032` (Regular) + its non-foil skuIds

## Root cause

Scryfall's `tcgplayer_id` for these cards points at the **Surge Foil** product (`698031`), not the **Regular** product (`698032`). `TcgplayerSkus` is built by joining flattened SKUs against the card's product-ID identifiers, so only the surge product's foil SKUs got attached; the regular product was referenced by nothing.

`_build_tcg_alt_foil_lookup` (`mtgjson5/data/context.py`) assumed the *non-suffixed* base product is always the card's anchor and the *suffixed* `(Surge Foil)` product is the alternative. That produced a `base -> alt` mapping (`698032 -> 698031`) which:
1. never joined — the card's `tcgplayerId` is the suffixed product (`698031`), not the base key (`698032`); and
2. was then deleted outright by the dedup step, because `698031` is a known Scryfall id.

Net result: `tcgplayerAlternativeFoilProductId` stayed null and the regular product was orphaned.

## Fix

Make the base ↔ alt-foil pairing **anchor-aware**: key the mapping on whichever product is a known Scryfall id and point it at the other product.

| Case | Mapping |
|---|---|
| Normal (Scryfall → base) | `base → alt-foil` (unchanged) |
| Flipped (Scryfall → Surge Foil, MSC) | `surge → regular` |
| Both products are distinct Scryfall cards | no mapping |

[...]

PR #1698, opened 2026-08-07
fix missing TCGplayer prices for alternative products

Emnalyeriar:
This fixes missing TCGplayer Market prices for cards that use `tcgplayerAlternativeFoilProductId`.

The alternative product mapping was already created and published, but the price builder only used the base and etched product mappings. As a result, prices attached to an alternative product were not included in `AllPrices.json` or `AllPricesToday.json`.

This follows:

- [#1545](https://github.com/mtgjson/mtgjson/pull/1545), which added the alternative product support.
- [#1678](https://github.com/mtgjson/mtgjson/pull/1678), which made the mapping anchor-aware and noted that alternative-product prices were still missing because the price builder did not use this mapping.

## Changes

- Use the alternative product mapping in both live and cached price builds.
- Load the mapping during standalone price builds.
- Determine `normal`, `foil`, or `etched` from TCGplayer's price data.
- Keep base and etched mappings authoritative when product IDs conflict.
- Avoid ambiguous mappings and duplicate price rows.

## Example

Jaws, Relentless Predator (`373dab28-3a38-5b34-abdb-f77d1801a6d2`) uses:

- Base product: `656542`
- Alternative product: `656544`

Product `656544` has a Foil Market price. This price is now published for the card.

### DIFF
--- a/mtgjson5/build/prices/price_builder.py
+++ b/mtgjson5/build/prices/price_builder.py
@@ -65,6 +65,7 @@ class PriceBuilderContext:
     Owns the ID -> UUID mappings needed by price providers:
     - tcg_to_uuid: TCGPlayer product ID -> MTGJSON UUIDs
     - tcg_etched_to_uuid: TCGPlayer etched product ID -> MTGJSON UUIDs
+    - tcg_alt_foil_to_uuid: Alternative TCGPlayer product ID -> MTGJSON UUID
     - mtgo_to_uuid: MTGO ID -> MTGJSON UUIDs
     - scryfall_to_uuid: Scryfall ID -> single MTGJSON UUID
 
@@ -76,6 +77,7 @@ class PriceBuilderContext:
     # Derived mappings (built lazily)
     _tcg_to_uuid: dict[str, set[str]] | None = field(default=None, repr=False)
     _tcg_etched_to_uuid: dict[str, set[str]] | None = field(default=None, repr=False)
+    _tcg_alt_foil_to_uuid: dict[str, set[str]] | None = field(default=None, repr=False)
     _mtgo_to_uuid: dict[str, set[str]] | None = field(default=None, repr=False)
     _scryfall_to_uuid: dict[str, str] | None = field(default=None, repr=False)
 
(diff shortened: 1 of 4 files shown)


the-muppet2:
Thanks for the PR, this closes the gap from #1678 nicely, and the test coverage is appreciated. A few things came up in review:

**1. Cross-group dedup ignores mapping priority in the live path (`prices.py`, `_finalize`)**

`_fetch_group_prices` enforces base > etched > alt precedence only within a single group's `candidates` dict. When a card's base and alternative products live in *different* TCGplayer groups (common for promo/extras products), both records land in `self._buffer`, and `_finalize`'s `unique(keep="first")` keeps whichever group happened to be fetched first, since the buffered records carry no priority field. So an alt-foil price from an earlier group can override the base price from a later group, which is the opposite of the "base and etched mappings stay authoritative" goal, and it diverges from the raw-cache path (`_map_tcg_raw`), which sorts globally by `_mappingPriority` before deduping. Could the buffered records carry the mapping priority so `_finalize` can sort by it before the `unique()`, matching the raw path?

**2. Lexicographic productId tie-breaking (both paths)**

In `_map_tcg_raw`, `productId` is cast to `pl.String` before `.sort(["_mappingPriority", "productId"]).unique(keep="first")`, so ties between two same-priority products are broken lexicographically: `'1000'` sorts before `'999'`. The live path's `rank = (mapping_priority, product_id)` has the same issue since `product_id` is a string there too. Probably worth sorting numerically so the winner is at least stable and predictable across releases.

**3. Possible etched-price loss from the base-beats-etched anti-join (`_map_tcg_raw`)**

The anti-join removes an etched mapping row whenever its productId also appears in `tcg_to_uuid`, even when the two mappings point at *different* UUIDs (product is card A's base ID and card B's etched ID). Previously both rows were emitted, so card B got its etched price; now it silently disappears. If the intent is only to resolve same-UUID conflicts, the anti-join may need to key on (productId, uuid) or only drop rows where the base mapping covers the same card.

**Smaller cleanups:**

- `GlobalCache.get_tcg_alt_foil_to_uuid_map` (cache.py) is added but never called; `_build_tcg_alt_foil_to_uuid` in price_builder.py hand-rolls the same collect/iter_rows/setdefault logic instead. Delegating to the cache accessor would keep one implementation.
- In `_fetch_group_prices`, `is_etched` is always exactly `mapping_priority == 1`; deriving the finish from the priority would avoid the two flags drifting apart if another mapping tier is ever added.
- `alt_foil_uuid_path` / the alt-foil map params default to `None`/`or {}` in several signatures, but every call site passes them. Making them required would let an accidentally-omitted argument fail loudly instead of silently dropping all alternative-product prices.
- The `.unique()` before the `group_by("productId").agg(...)` in the alt-foil mapping pipeline is redundant; `n_unique()` and `.first()` give the same result without it.

Emnalyeriar:
@the-muppet2 thanks for the code review and comments! I applied the necessary changes :)

Issue #1650, opened 2026-05-03
keep uuid constant for sealed and tokens


kodawah:
for sealed, right now if a name changes for whatever reason, the uuid changes as well

as for tokens, i dont know what's happening

PR #1712, opened 2026-09-03
Keep sealed product UUIDs stable across renames

ZeldaZach:
Closes half of #1650 (the sealed half). Tokens will follow in a separate PR.

## The problem

Sealed product UUIDs are `uuid5(NAMESPACE_DNS, productName)` and nothing else, computed independently in three places. The name *is* the identity, so renaming a product in mtg-sealed-content mints a brand new UUID and breaks every reference to it:

- `sealedProduct[].uuid`
- `contents.sealed[].uuid`
- `card.sourceProducts`
- the hashed purchase URLs

This is not hypothetical. 273 distinct top-level product keys were removed from `mtg-sealed-content/data/products` across 300 commits since January 2025, and a good share are renames rather than deletions. One recent example flipped twice in a week: `Star Trek Commander Deck Federation Fleet` → `Star Trek Collector Commander Deck Federation Fleet` and back (`d4b55dc`, `8703052`, `3bba4a4`).

## The fix

A pin file, `mtgjson5/resources/sealed_product_uuids.json`, records the UUID each product was first published with plus the identifiers it carried at the time. Resolution order per product:

1. Exact `setCode` + name hit → reuse the pin.
2. Otherwise match against *retired* pins (ones whose name is gone from the current build) by shared identifier → reuse, treating it as a rename.
3. Otherwise mint `uuid5(name)`, exactly as today.

Rename matching is deliberately conservative. A pin is reused only when the pairing is unambiguous **in both directions**: the product matches exactly one retired pin, and that pin is claimed by exactly one product. That two-way check is what stops a newly added sibling from stealing a live product's UUID, which matters because `Case` variants routinely share identifiers with the base product. Anything ambiguous logs a warning and falls through to a fresh UUID.

Coverage is good: of 4,209 current products, 4,034 (95.8%) carry at least one of tcgplayerProductId / cardKingdomId / mcmId / cardtraderId / csiId / abuId / tntId / scgId, and there are zero cases where one `(setCode, identifier, value)` maps to two names.

The three duplicated `uuid5` call sites now share one resolver in `mtgjson5/pipeline/stages/sealed_uuids.py`.

## Why an expression and not a join

The `sealedProduct` array order in the output is the row order of the frame built in `build_sealed_products_lf`, and nothing sorts it downstream — so row order is user-visible. Resolution therefore goes through a `replace_strict` mapping expression, which cannot reorder rows, rather than a left join. (`maintain_order="left"` was tried and does not actually hold order under the streaming engine.)

## This changes no existing output

The seeded pin file reproduces all 4,209 currently published UUIDs exactly — every pin equals `uuid5(name)` today, with zero collisions. Verified end to end by running `build_sealed_products_lf` against the real build cache:

- 4,209 rows, no join duplication, no null UUIDs
- BRO's 23 sealed products match the published build in both **UUID and array order**
- output deterministic across repeated runs

Then, renaming `The Brothers' War Bundle` in the real parquet and rebuilding:

```
renamed product uuid : 6d0880d3-b3a8-5df5-a526-5be1def7899b
original (pinned)    : 6d0880d3-b3a8-5df5-a526-5be1def7899b
would have been      : 16b66ea7-535b-5a26-bd0b-c2c1c2e73157
```

A pin carried through a rename records `originalName`, the name its UUID was minted from, so the whole file stays auditable against the historical formula even after products move around. A test enforces that invariant.

## Maintenance

`scripts/update_sealed_uuid_pins.py` refreshes the pin file as products land upstream, with `--check` for CI. It reads from a local checkout, the build cache, or GitHub, and warns when the cache it picked is over a day old.

I did **not** wire `--check` into `tests.yml`: CI has no populated build cache, so it would download the mtg-sealed-content tarball on every run. That felt like flakiness worth opting into deliberately rather than inheriting — happy to add it, or set it up as a scheduled job instead.

## Notes

- `.gitignore` has a blanket `*.json` with a single negation for `mkm_finish_profiles.json`; added a second negation for the pin file. The other tracked resource JSONs predate that rule.
- Pins are scoped per set code, so a product moving between set files would mint a new UUID. No such cases exist in current data; a cross-set fallback is easy to add if it comes up.

## Testing

947 tests pass (23 new in `tests/mtgjson5/test_sealed_uuids.py`), plus ruff, ruff-format, mypy, and pylint 10/10.

https://claude.ai/code/session_01K3qgAgGDiPNKDPetCyktRP


kodawah:
Ran a deep review of this PR, with the mechanisms below verified by executing repros against the PR head and the real build cache. First, what checks out: resolution is byte-identical to base on current data (all 4,209/4,210 products; the one extra is the SLZ pack that postdates the pin file and falls back identically), `polars_hash`'s `uuidhash.uuid5()` matches Python `uuid5(NAMESPACE_DNS, ...)` byte-for-byte including non-ASCII names, all three mint sites route through the resolver (AllPrintings streaming, purchase-URL hashing, deck links, and `sourceProducts` all flow from it), and the row-order reasoning for using an expression instead of a join is sound.

The findings are almost all about what happens after merge when the data moves. A common thread: every guard degrades to a `LOGGER.error`/`warning`, so each of these ships silently.

**1. Reused retired name ⇒ two live products publish one UUID.** After a rename carries a pin forward, upstream re-adding a product under the old name gives both products `uuid5(oldName)` (reproduced). `_warn_on_collisions` (`sealed_uuids.py:219`) only logs; `sealedProduct[]` then carries two entries with one uuid, and `_ctp_get_cards_in_sealed_product` matches by uuid and `break`s, so `sourceProducts` attributes one product's contents to both. Pre-PR, a rename minted a fresh UUID, so this duplicate channel is new. Same collision fires cross-set, since the name formula ignores set code.

**2. Pin edits mid-cache-window ⇒ internally inconsistent release.** `sealedProduct[].uuid` is re-resolved from the live pin file every build (`metadata.py:355`), but `contents.sealed[].uuid`, deck `sealedProductUuids`, and `card_to_products` are baked into the github parquets for up to 290h (`_cache_fresh`), and the `all_cached` fast path skips `on_github_complete` entirely. Nothing keys those caches on the pin file, so a resolution-changing pin edit (an ambiguity fix, a manual correction — exactly the file's purpose) landing inside the window ships one AllPrintings where the two sides disagree for the same product. Folding the pin file's hash/mtime into the derived parquet cache keys closes it.

**3. The resolver's call sites see different product universes.** Resolution is context-sensitive (present vs vacated names), but `metadata.py` feeds it `ctx.sealed_products_lf` — whose provider drops products with malformed non-dict YAML bodies — while `sealed.py` feeds raw `compile_products()` output, which keeps them. Reproduced: with a malformed entry whose pin shares an identifier with a new product, the frame path vacates the pin and rename-matches it while the dict path holds it, so the same product gets two different UUIDs within one build. In base, uuid was a pure function of name, so the sites could never disagree.

**4. One shared identifier suffices for a rename match.** `_match_renames` unions candidates over any single `(field, value)` hit (`sealed_uuids.py:123`) and never treats conflicting identifiers as counter-evidence. A new product carrying one vendor-reused or mistyped id can pass the two-way test "unambiguously" and inherit a retired product's published UUID (reproduced). The exposure is real: the shipped pin file already contains same-set duplicate identifier values (ACR `tntId` 1808837 on Starter Kit + its Case, HOB `mcmId` 885554, KTK `tntId` 1182106), and retired pins are kept forever. Requiring no conflicting fields (or ≥2 agreeing ones) would harden it.

**5. `build_pins` can delete a live product's pin.** The refile loop treats any same-UUID pin as a vacated rename entry; the `old_name in written` guard (`sealed_uuids.py:260` area) only protects names already processed this run. With the finding-1 duplicate state, when the name-reusing product iterates first, the live product's pin is deleted and re-seeded without `originalName` — the regenerated file then fails this PR's own `test_matches_the_published_name_based_uuids` (reproduced). The guard wants `old_name not in set_products`, not just `not in written`.

**6. Cross-set name reuse writes duplicate pins that fail the shipped-file test forever.** Retired pins are kept, dedup only scans the current set, and the name formula ignores set code — so a product re-added under another set code lands the same uuid under two sets (reproduced). No warning fires anywhere (`_warn_on_collisions` only sees live products), `--check` converges and reports "up to date", and `test_no_uuid_is_shared_by_two_products` stays red until hand-edited. This is the concrete cost of the per-set scoping the PR body flags; a global second pass with the same two-way guard would subsume it.

**7. Name match overrides identifier evidence.** Resolution matches by current name before identifiers are consulted (`sealed_uuids.py:191`), so a simultaneous name swap resolves each product to the other's UUID with empty logs — and `build_pins` then permanently refiles the swapped identifiers under the wrong pins, poisoning future rename detection (reproduced). The swap itself matches base `uuid5(name)` behavior, but the module holds the evidence to detect it and instead corrupts that evidence. Comparing a pin's stored identifiers against the build's for the same name would at least warn.

**8. `--check` compares parsed dicts, not bytes — and trusts a stale cache.** `update_sealed_uuid_pins.py:124` re-serializes both sides, so a hand edit, merge reformat, or duplicate JSON key (silently last-wins in `json.loads`) passes CI while the next real run produces a whole-file diff; #1713's twin script gets this right with a byte compare against `serialize_pins`. Separately, `_resolve_products_dir` prefers a possibly-days-stale cache with only a log line, so check mode can validate against data predating the very rename being checked.

**9. Latent: `sealed_uuid_expr`'s key skips uppercasing.** `resolve_sealed_uuids` uppercases set codes (docstring: "may be any case") but the lookup key concatenates the raw `setCode` column (`sealed_uuids.py:316`); a non-uppercase row resolves to null and the `pl.coalesce` fallback silently mints the name UUID — pins quietly disabled. Harmless today (verified the provider uppercases and 4,210/4,210 rows resolve), but `str.to_uppercase()` in the key — and making the otherwise-unreachable fallback assert instead of coalesce — turns a future mismatch loud.

Cross-PR notes: this PR and #1713 insert their `.gitignore` negation at the same line — whichever merges second needs a trivial rebase (verified conflict). And the two ship near-identical pin-store scaffolding (`load_pins`/`dump_pins`/`--check`) that already diverges on indent and check semantics; with #1650 implying more pin files to come, a small shared module would keep them from drifting.

ZeldaZach:
Worked through the review — 9c91c03 fixes eight of the nine, and the ninth is a note rather than a change. Resolution over the real build cache is still unchanged: 4,209 products, every one matching its published name-based UUID, no duplicates, and `build_pins` regenerates the shipped file byte-identical.

**1 + 6 (duplicate UUIDs after a name is reused, same set or across sets)** — confirmed, and the cross-set half is the same bug since the formula ignores set codes. Resolution now ends in `_split_collisions`: the pinned product keeps the UUID it was published with and the newcomer is re-minted from `"{SET} {name}"`, recorded as its `originalName` so the pin still equals `uuid5(originalName)` and `test_matches_the_published_name_based_uuids` keeps holding. A collision with no pin behind it — two sets sharing a product name, which the base formula collides on too — is only reported, since re-minting there would move a UUID already in the wild.

**2 (pin edit inside the cache window)** — fixed as suggested. `pin_file_digest()` is written to `github_sealed_pins.sha256` next to the derived parquets and checked before the `all_cached` fast path, so a pin edit recompiles the sides that bake UUIDs in rather than letting them disagree with `sealedProduct[]`.

**3 (call sites see different product universes)** — good catch, and the reproduction holds. Both paths now go through `_usable_products`, which drops non-mapping product bodies and malformed set bodies the same way `_build_sealed_products_records` does, so the two callers vacate the same pins.

**4 (one shared identifier is enough)** — `_match_renames` now drops a candidate that disagrees on any field the two both carry. Requiring ≥2 agreeing fields would have cost real renames on the 4.2% of products carrying a single id, so contradiction-as-counter-evidence seemed the better trade.

**5 (`build_pins` deletes a live pin)** — the refile guard is `old_name in usable` now, not `old_name in written`.

**7 (name match overrides identifier evidence)** — not changing the resolution order (matching base `uuid5(name)` behaviour is the point), but a pinned product whose identifiers disagree with its pin now warns with the specific fields, so a swap leaves a trail instead of silently poisoning the pins on the next refresh. Zero products in the current data trip it.

**8 (`--check` compares dicts and trusts a stale cache)** — now compares `serialize_pins(updated)` against the file's bytes, so a hand edit, merge reformat or duplicate JSON key fails the way it should. In check mode a build cache older than 24h is ignored in favour of a fresh download rather than validating against data that predates the rename being checked.

**9 (`sealed_uuid_expr` key case)** — `str.to_uppercase()` on both sides of the lookup. Kept the `pl.coalesce` fallback: it is unreachable for rows coming out of `products_lf`, but the expression is public enough that a null-safe tail is worth more than an assert here.

On the cross-PR notes: the `.gitignore` negation still conflicts with #1713 and whichever lands second gets the one-line rebase. Leaving the shared pin-store module for a follow-up — with #1650's token half still to come it is worth doing once against three call sites rather than twice against two, and folding it in now would put a refactor of #1713's file in this PR.

947 → 961 tests pass, plus ruff, ruff-format, mypy, and pylint 10/10.

Issue #1644, opened 2026-04-18
token UUIDs change often


kodawah:
not sure what's up, but there seems to be something that causes instability of on the token uuids

```diff
@@ -485110,7 +485190,7 @@
           "rarity": "mythic",
           "relatedCards": {
             "tokens": [
-              "2a43d4ad-81b6-5f95-b42f-db833ded0112"
+              "ebb9c0a4-c359-58de-b739-03fddd775dfa"
             ]
           },
           "rulings": [
@@ -489753,7 +489833,7 @@
           "rarity": "rare",
           "relatedCards": {
             "tokens": [
-              "2a43d4ad-81b6-5f95-b42f-db833ded0112"
+              "ebb9c0a4-c359-58de-b739-03fddd775dfa"
             ]
           },
           "rulings": [
@@ -489871,7 +489951,7 @@
           "rarity": "rare",
           "relatedCards": {
             "tokens": [
-              "2a43d4ad-81b6-5f95-b42f-db833ded0112"
+              "ebb9c0a4-c359-58de-b739-03fddd775dfa"
             ]
           },
           "rulings": [
@@ -536435,7 +536515,7 @@
           "rarity": "mythic",
           "relatedCards": {
             "tokens": [
-              "2a43d4ad-81b6-5f95-b42f-db833ded0112"
+              "ebb9c0a4-c359-58de-b739-03fddd775dfa"
             ]
           },
           "rulings": [
```

in particular it seems like db833ded0112 got replaced with 03fddd775dfa, but i can't find db833ded0112...

kodawah:
note: this particular case affected relatedCards.tokens, but this kind of instability happens for the tokenparts as well

kodawah:
merged in #1650

Issue #1591, opened 2026-03-07
token uuids keep changing


kodawah:
Not sure what's going but some of the token uuids are not stable, at least in the relatedCards section

```diff
--- allprintings5-OLD.json.noindex	2026-03-03 17:55:42
+++ allprintings5.json	2026-03-07 11:33:18
@@ -1,7 +1,7 @@
 {
   "meta": {
-    "date": "2026-03-03",
-    "version": "5.3.0+20260303"
+    "date": "2026-03-06",
+    "version": "5.3.0+20260306"
   },
   "data": {
     "10E": {
@@ -7933,7 +7933,7 @@
           "rarity": "rare",
           "relatedCards": {
             "tokens": [
-              "5aea21ce-bd08-5cd4-87f1-991f8f708d7c"
+              "57cf269e-c2d8-5b57-9436-5659a93200cc"
             ]
           },
           "setCode": "10E",
@@ -8023,7 +8023,7 @@
           "rarity": "rare",
           "relatedCards": {
             "tokens": [
-              "5aea21ce-bd08-5cd4-87f1-991f8f708d7c"
+              "57cf269e-c2d8-5b57-9436-5659a93200cc"
             ]
           },
           "setCode": "10E",
@@ -40966,7 +40966,7 @@
           "rarity": "rare",
           "relatedCards": {
             "tokens": [
-              "63ac47ed-3457-5af8-9a33-1bb5dcb9ca9f"
+              "22f55141-b82d-57c7-bbed-4804cf6ea23a"
             ]
           },
           "setCode": "10E",
@@ -50396,7 +50396,7 @@
           "rarity": "rare",
           "relatedCards": {
             "tokens": [
-              "7decf258-eb10-50da-83f7-c7eba74adbfb"
+              "2b47c8df-c600-5d6c-8d52-3ffd64a1c629"
             ]
           },
           "setCode": "10E",
@@ -50486,7 +50486,7 @@
           "rarity": "rare",
           "relatedCards": {
             "tokens": [
-              "7decf258-eb10-50da-83f7-c7eba74adbfb"
+              "2b47c8df-c600-5d6c-8d52-3ffd64a1c629"
             ]
           },
           "setCode": "10E",
@@ -59731,7 +59731,7 @@
           "rarity": "rare",
           "relatedCards": {
             "tokens": [
-              "8b44e4ca-fb3a-5ef1-a607-6ed5b98cb216"
[...]

PR #1713, opened 2026-09-03
Keep relatedCards.tokens stable across Scryfall repoints

ZeldaZach:
Closes the token half of #1650, which also absorbed #1591 and #1644. The sealed half (#1712) has merged; this is rebased on it.

## The problem

Token UUIDs themselves are fine. The churn is in `relatedCards.tokens`.

A card's related tokens come from Scryfall's `all_parts`, which names one specific *printing* of each token. For a token the card's own set never printed — Copy, Treasure, Clue — Scryfall picks an arbitrary printing and re-picks it over time. `add_token_ids` follows that pointer faithfully, so the UUID changes even though nothing about the card or the token did.

The Copy token from #1644 has pointed at three different printings:

| UUID in the diffs | printing |
|---|---|
| `2a43d4ad…` | Copy, `ttmt` #1 |
| `ebb9c0a4…` | Copy, `tsos` #1 |
| `3ae3c4af…` | Copy, `tmsc` #17 (today) |

6,553 of 16,328 English token references point outside the card's own set and are exposed to this. Concentrated in Copy (826), Treasure (571), Zombie (287), Spirit (271), Clue (179).

This is live, not historical: comparing the current Scryfall dump against a local May build, 2X2's Twinflame and Bloodforged Battle-Axe have already drifted from `ebb9c0a4` to `3ae3c4af` since then.

## The fix

`mtgjson5/resources/token_references.json` records the printing each card was first published against:

```
{"<card scryfall id>": {"<token oracle id>": "<token scryfall id>"}}
```

Keyed on the token's **oracle id**, the identity that survives a printing swap. Name alone would not do: 66 cards relate to multiple same-named tokens (Golem being the usual culprit), and name + type_line still leaves 61 ambiguous.

The pin stores the **scryfallId**, not the resolved UUID, so it flows through the existing `scryfall_uuid_lf` lookup and legacy UUID-cache pins still apply. A pin whose printing has left Scryfall resolves to null and falls back to the live reference, so a pin can never strand a token.

Liveness is computed from every id in the dump, deliberately not from the ids still referenced. A repoint is precisely the case where the old printing stops being referenced while still existing — deriving liveness from the references would retire every pin the moment it did its job.

### What the file covers

19,302 raw references in the dump reduce to 16,017 pins across 15,228 cards (1.9 MB):

- **3,285 dropped** as unreachable: `signatures.py` nulls `relatedCards.tokens` for token-layout cards, so pinning their references only padded the file.
- **3 added** through the face-level oracle: Scryfall gives `reversible_card` printings no top-level `oracle_id`, so the SLD Mechtitan token's references had no key to file under and stayed exposed to the exact churn this PR exists to stop. `oracle_id_expr()` coalesces the top-level id with the first face's, which is enough to cover every printing in the dump (0 null oracles across 541,398).
- A card naming **two printings of one oracle** is left unpinned rather than collapsed. Zero exist today, but the `(card, oracle)` key cannot represent one, and pinning either printing would drop the other from published output.

## This changes no existing output

Verified on the frame the pipeline actually receives (`GlobalCache` normalizes the dump to camelCase, so it carries `oracleId`/`allParts`/`cardFaces.oracleId`): across all **21,811** English cards with token references, pinned and unpinned resolution are identical.

Then, simulating a Scryfall repoint of all 38 Copy printings onto one:

```
repoint, unpinned == baseline: False
repoint, pinned   == baseline: True    (18,854 publishing cards)
```

The committed pin file reproduces byte-identically from the current dump, from scratch and as a merge.

## Note on what pins preserve

Pins seeded from today's dump freeze *today's* values. They do not retroactively restore what an older build published — the original values aren't recoverable without historical data, and today's value is what the next build emits regardless. If anchoring to the currently published AllPrintings is preferable, that's a small change to the generator.

## Cost

The pins ride along with work the build already does. `_build_card_identity_df` collects `id`/face-count/`oracleId` in one scan and hands it to both the global UUID map and the pin builder, so the pipeline parses the 2.9 GB dump once where it previously would have twice. Pinned printings are resolved to UUIDs once in the pre-pass rather than joined against the million-row UUID map inside every batch.

## Maintenance

`scripts/update_token_reference_pins.py` refreshes the pin file from the cached Scryfall dump, with `--check` for CI. It aborts when more than `--max-repoints` (default 25) pins would move to a different printing: a healthy refresh repoints nothing, so any volume means the dump is partial rather than that Scryfall retired that many printings — which would otherwise rewrite thousands of pins in one silent 55k-line diff. It also samples the schema as deeply as `GlobalCache` does, so both sides agree on whether the face-level oracle fallback is available.

As with #1712 I did not wire `--check` into `tests.yml` — CI has no populated cache, and this one needs the 2.9 GB dump, so it wants to be a scheduled job rather than a per-PR check.

## Known limitation

When the printing `all_parts` names today is missing from the dump entirely, there is no oracle id to reach the pin through, so the reference resolves to nothing — exactly as it did before pins existed. Zero cases in the current dump; documented in the module docstring rather than papered over with a heuristic that would have to guess which pin a dangling pointer meant.

## Testing

996 tests pass (28 in `tests/mtgjson5/test_token_references.py`, covering the face-oracle fallback, the token-layout skip, the ambiguous-oracle case, and a camelCase end-to-end case), plus ruff, ruff-format, mypy, and pylint 10/10.

https://claude.ai/code/session_01K3qgAgGDiPNKDPetCyktRP


kodawah:
Ran a deep review of this PR, with the mechanisms below verified by executing repros against the PR head and the real 2.9 GB dump. What checks out: the shipped pin file reproduces byte-for-byte from the current dump (19,299 refs / 18,032 cards, `--check` green), pinned vs unpinned resolution is identical on current data, the camelCase/snake_case dual handling matches what the pipeline and script actually see, null join keys can't spuriously match pins, and the liveness-from-the-whole-dump reasoning is correct. One candidate finding — that pinning only the forward direction desyncs `reverseRelated` — was checked against raw Scryfall records and refuted: all printings of a token share the same related-parts group, so no published mismatch can occur.

The findings:

**1. `reversible_card` token references are silently unpinnable — 3 live today.** Scryfall gives reversible printings no top-level `oracle_id`, so `extract_token_references` drops their references (`token_references.py:140`; raw 19,302 → 19,299 kept) and `build_token_pins` can never apply even a hand-added pin (a null oracle never matches). Concretely: Mechtitan Core (prm, sld) and the SLD reversible card → the SLD "Mechtitan // Mechtitan" reversible token. When Scryfall repoints those, `relatedCards.tokens` churns — the exact #1644 symptom — with zero logging. Falling back to the face-level `oracle_id` would cover them.

**2. Latent: the (card, oracle) key collapses two printings of one oracle.** A card whose `all_parts` legitimately names two printings of the same token oracle has both rows join the same pin, and the new `.unique()` agg collapses them to one entry — dropping a published reference relative to base (reproduced: `[uuid(P1)]` vs `[uuid(P1), uuid(P2)]`). Zero such cards exist in today's dump, but the dedup comment in `extract_token_references` anticipates the shape, the pin layout can't represent it, and when it occurs the loss is permanent and unlogged.

**3. The update script mass-repoints on a partial dump, unguarded.** `build_pins` auto-repoints any pin absent from `live_printings` with no volume guard, and the script checks nothing beyond `exists()` and `height > 0`. Mitigating: polars raises `ComputeError` on a torn line, so a mid-line truncation fails loudly. But `_convert_file_to_ndjson` writes in place (no tmp+rename), a record-boundary truncation parses cleanly, and `--cards default_cards.ndjson` is a guaranteed silent path — exit 0, INFO logs, thousands of pins rewritten in one 55k-line diff. An abort threshold on repoint volume (and atomic conversion) would make this safe.

**4. A dangling live pointer bypasses a valid pin.** Pins are reached only through the live printing's oracle id, so when Scryfall's current `all_parts` id is absent from the dump (a case your own tests cover), the pin join never fires and the reference drops to `[]` even though the pin records a resolvable printing (reproduced). Output matches base, but the pin layer holds exactly the information needed to preserve the reference through this churn — a fallback join on the card's own pin set would use it.

**5. 13% of the shipped pin file is inert self-pins.** 2,529 of 19,299 entries pin a token to itself (plus 316 token→token refs) — all token/double_faced_token/flip-Role layouts, and `signatures.py` nulls `relatedCards.tokens` for token layouts, so none can ever reach output (verified end to end). The refresh script adds one permanent dead pin per newly printed token. Filtering token-layout cards (or `_card_sid == _tok_sid`) at extraction shrinks the 2.3 MB resource and its diff noise.

**6. Efficiency: the pre-pass re-parses the full dump; every batch re-joins the big lookup.** `build_token_pins` collects `oracle_lf` via its own scan of `ctx.cards_lf` right after `_build_global_scryfall_uuid_map` collected a narrow scan of the same file — separate `.collect()`s never share a scan, so every build (single-set included) pays an extra full 2.9 GB parse; the script likewise double-parses. And `add_token_ids` joins the ~0.5–1M-row uuid lookup on `_pin_sid` inside every batch for a fixed 19k mapping. Widening the existing pre-pass to also select `oracleId` (or `pl.collect_all`) and pre-resolving `_pin_sid → uuid` once would remove both.

Cross-PR note: this PR and #1712 insert their `.gitignore` negation at the same line — whichever merges second needs a trivial rebase (verified conflict).

ZeldaZach:
Thanks — all six reproduced against the 2.9 GB dump. Five are fixed in 930ea23; one I deliberately left alone.

**1. Reversible references — fixed.** Confirmed exactly 3, all pointing at the SLD `Mechtitan // Mechtitan` token, and the face-level `oracle_id` is there (`a4fecf0a…`, identical on both faces). `oracle_id_expr()` now coalesces the top-level id with the first face's. One wrinkle worth recording: the field is only *visible* at sufficient schema-inference depth — `card_faces` has no `oracle_id` at `infer_schema_length=2000`, which is what the script used, but does at the 100k `GlobalCache` uses. Generator and pipeline disagreeing there would have filed pins under one key and looked them up by another, so the script now samples at 100k too and `_face_oracle_expr` returns None rather than raising when the field is absent. Identity frame is now 0 null oracles across 541,398 printings.

**2. Two printings of one oracle — fixed.** Left unpinned instead of collapsed: the group is dropped at extraction with a warning, so both references follow Scryfall and both survive to output. Confirmed 0 such groups today, and there's a test that both UUIDs come back for the pair.

**3. Unguarded mass repoint — fixed.** `build_pins` now returns a `PinMerge` carrying the counts, and the script aborts before writing when `repointed > --max-repoints` (default 25). One correction to the finding: `--cards default_cards.ndjson` does *not* repoint anything today — 0 repointed, verified — nor does a 150k-line truncation. A repoint needs the pin to differ from the live reference *and* the pinned printing to be absent, and since every pin currently equals its live value, the pinned printing is present by construction. The exposure is real but it's a future one, arriving the first time pins diverge; the guard fires correctly on a synthetic 30-pin diverged case (exit 1, nothing written). I did not touch `_convert_file_to_ndjson` — the atomic-write fix belongs with the download path, not here.

**4. Dangling live pointer — not fixed, deliberately.** Zero cases in the dump (the 3 unpinnable references were all present-but-null-oracle, not absent). Output matches base, and reaching the pin would mean guessing which of a card's pins a dangling pointer meant — the oracle is exactly what's unavailable. Recorded in the module docstring instead. Happy to revisit if you see a non-heuristic route.

**5. Inert self-pins — fixed.** Filtered on token layout rather than `_card_sid == _tok_sid`, which removes 3,285 references (all 2,529 self-pins plus the 316 token→token and the rest) against 2,845 for the narrower rule. Checked that `layout` isn't rewritten before `signatures.py` reads it — the only rewrite in the pipeline is split→aftermath — and that `_token_uuids` feeds nothing but `relatedCards` (`propagate_salt_to_tokens` goes through `reverseRelated`). 2.3 MB → 1.9 MB. Six self-pins survive on non-token layouts; left them, since they're inside the set the layout rule can't justify dropping.

**6. Double parse and per-batch join — fixed.** New `_build_card_identity_df` collects `id`/`_n_faces`/`oracleId` in one pass and feeds both the UUID map and the pins, so the pipeline parses the dump once. `build_token_pins` now takes `scryfall_uuid_lf` and pre-resolves pins to UUIDs, so `TokenPins.pins_lf` carries `_pinned_uuid` and `add_token_ids` drops that per-batch join entirely. The script gets `scan_references_and_printings`, which runs both scans through one `collect_all`. I made `scryfall_uuid_lf` a required argument rather than optional — defaulting it would have quietly put Scryfall ids in the slot that gets coalesced as a UUID.

**Cross-PR note:** rebased on master now that #1712 has merged; the `.gitignore` negation carries both lines.

Regenerated pin file: 19,299 → 16,017 references over 15,228 cards. Every retained pin is value-identical to the previous file (3,285 removed, 3 added, 0 changed). Re-verified end to end on the camelCase frame — pinned == unpinned across all 21,811 English cards with token references, and under a simulated repoint of all 38 Copy printings the pinned result holds all 18,854 publishing cards steady while the unpinned one churns. 996 tests, ruff, mypy, pylint 10/10.

PR #1728, opened 2026-09-16
data/context: fall back foreignData name to the card's own name

kodawah:
## Bug

`_build_foreign_data_df()` drops any non-English printing whose Scryfall object has no transcribed `printed_name`, via `.filter(pl.col("_foreign_name").is_not_null())`.

Confirmed against the real Scryfall bulk data: all four cards in "Secret Lair Drop Special Guest Junji Ito" (SLD 1114-1117) have genuine Japanese printings — their own distinct Scryfall IDs — but `printed_name`/`printed_text`/`printed_type_line` are all null for them. So their `foreignData` entries were silently dropped, even though a Japanese object legitimately exists and mtgban.com users can see it's missing (see mtgjson/mtg-sealed-content discussion — the Japanese and English "Etched" SLD SKUs currently show identical decklists).

## Fix

The multi-face branch already falls back per-face:
```python
pl.coalesce(pl.element().struct.field("printed_name"), pl.element().struct.field("name"))
```
but the single-face branch (`.otherwise(pl.col("printedName"))`) had no equivalent fallback. This brings it in line: fall back to the card's own (English/oracle) `name` when the localized one is missing.

The computed UUID (`uuid5(defaultScryfallId + side + "_" + language, NAMESPACE_DNS)`, added in #1727) never depended on the name at all — only this filter did — so nothing else about the row's computation changes; it simply survives instead of being dropped.

`text`/`type` are deliberately left null when untranslated rather than fabricated from the English text, so a `foreignData` entry doesn't imply we have real translated rules text when we don't.

## Verification

- New tests (`tests/mtgjson5/test_foreign_data_name_fallback.py`, 4 cases): null `printed_name` falls back to the card name, the fallback doesn't change the computed UUID (matches the real, independently-computed value for Carrion Feeder/SLD:1114), a genuine `printed_name` still wins over the fallback, and `text`/`type` stay null rather than being fabricated.
- Verified against the real Carrion Feeder / SLD:1114 case directly: the entry now resolves with name "Carrion Feeder" (fallback) and the exact UUID the deck-level language resolution would independently compute for it.
- A repo-wide check across every non-English deck-referencing sealed product in mtg-sealed-content found **12 of 18** affected by this same gap (Pictures of the Floating World, both Hatsune Miku drops, Junji Ito, Yoji Shinkawa, The Tokyo Lands) — this fix unblocks all of them, not just Junji Ito.
- Full existing suite: green.
- `ruff check`, `ruff format --check`, `pylint` (10.00/10), `mypy --explicit-package-bases` all clean on both changed files.

## Related

Companion to #1727 (language-aware `card:` resolution). A separate change is coming to extend the same language-aware resolution to `deck:`-referencing sealed products (Junji Ito's SLD products use `deck:`, not `card:`).


kodawah:
this one is good to go! hostile review

Also a clear ship — smaller and cleaner than #1729, and it’s the prerequisite for that one to matter on Junji Ito.

### What’s right
One real bug: `_build_foreign_data_df` dropped non-English printings with null Scryfall `printed_name`, even when a distinct Scryfall object exists (SLD 1114–1117 JA). The multi-face branch already `coalesce(printed_name, name)`; the single-face `.otherwise(printedName)` didn’t. Aligning them is the obvious fix.

UUID never depended on the name (`uuid5(defaultScryfallId + side + "_" + language)`), so keeping the row doesn’t change identity — only survival past `is_not_null()`. Leaving `text`/`type` null instead of copying English rules text is the right honesty rule.

Tests are tight: real Carrion Feeder IDs, UUID pin against the independent formula, genuine `printed_name` still wins, text/type stay null. Blast radius callout (12/18 non-English deck products) is useful.

### Soft nits
1. **Display semantics** — `foreignData[].name` can now be English under `language: Japanese`. Correct for “entry exists,” surprising for UIs that treat that field as “localized name.” One schema/docs sentence would save the next consumer an hour.
2. **Merge order** — land **#1728 before #1729**. Without this, language-aware `deck:` remapping still has nothing to remap to for those JA printings.
3. **Edge** — if both `printedName` and `name` were null, the row still drops. Fine for real Scryfall cards; not worth defending further.

**Verdict:** approve. No blockers. Pair with #1729 in that order and the mtgban Junji Ito EN/JA decklist collision has both halves of the data fix. Want a pasted review on the PR?

PR #1729, opened 2026-09-16
Add sourceProducts to foreignData and language-aware deck card_to_products

kodawah:
## Summary

Two related fixes to sealed-product/language handling, split out from investigating why [mtgban.com's Japanese vs. English Junji Ito Secret Lair decklist search](https://mtgban.com/search?q=decklist%3a%22Secret+Lair+Drop+Special+Guest+Junji+Ito+Japanese+Etched%22) returned identical results:

1. **`foreignData[]` entries had no `sourceProducts` field.** `add_source_products` only ever joined `card_to_products` against the outer card's own `uuid`, never against `foreignData[].uuid` — even though each foreign-language entry is its own distinct printing with its own uuid (see `_build_foreign_data_df`). This adds a `sourceProducts` field to the `ForeignData` schema (documented consistently with the existing `SkuIds` multi-parent pattern) and populates it per-entry in a new `_add_foreign_data_source_products`, so consumers like mtgmatcher can look up which sealed products a *specific foreign printing* is found in, not just the card's default-language one. This field is not theoretical — mtgban-website's `cardobject2sources` (`searchfilter.go:400`) reads a `CardObject`'s `SourceProducts` directly to answer several `decklist:`/`contents:`/sealed-status filters.

2. **`card_to_products` ignored a sealed product's own language when resolving `deck:` content.** A deck object (`mainBoard`/etc.) is shared verbatim by name across every sealed product that references it — its card uuids are always the deck's own default-language (English) printings. A language-tagged product (e.g. a Japanese-language reprint of an English-authored preconstructed deck) was attributed to the exact same English uuids as its untagged counterpart, because nothing in the deck-resolution path ever consulted the requesting product's language. `_ctp_get_cards_in_deck` now takes the enclosing product's `language` and substitutes each deck card's uuid for its `foreignData`-derived counterpart via the `cards_by_language`/`tokens_by_language` index (built in `build_uuid_map_from_pipeline`, from #1727), falling back to the default uuid with a warning when no foreign printing exists yet for that exact card.

Only `deck:` content consults the product's language; `card:` entries already carry their own independent `language` field (#1727).

## Design note / follow-up findings

While scoping the actual go-mtgban side of this, a few things worth recording here:

- **Items 1 and 2 are not independent — item 2 is a prerequisite for item 1 to do anything for deck-referencing products.** A foreignData uuid (e.g. the Japanese Carrion Feeder in Junji Ito) is never a row in the main cards frame, only nested inside `foreignData[]`. `_add_foreign_data_source_products` finds its sourceProducts by joining against `card_to_products_lf` on that nested uuid — and *that* frame only contains the Japanese uuid as a key at all because `compile_card_to_products`'s deck traversal now emits it (item 2). Without item 2, item 1 would silently produce empty `sourceProducts` for every deck-referencing foreign printing, i.e. exactly the Junji Ito case that motivated this PR. So these should land together, not split.
- **go-mtgban's `decklist:` search filter (the original symptom) calls `mtgmatcher.GetDecklist` directly** (`mtgban-website/searchfilter.go:773`,`790`), which walks `Set.Decks`/`SealedProduct.Contents` itself and never reads `sourceProducts`/`card_to_products`. So this PR does not, by itself, fix the mtgban.com symptom — that fix is being designed directly in go-mtgban: check the product's own language, look up the deck card's foreign-language uuid, and use it only if go-mtgban already has that specific foreign printing loaded/minted, otherwise keep the English uuid unchanged (no inventing unminted uuids). `sourceProducts`/`card_to_products` remains valuable independently, for the `cardobject2sources`-driven filters above.
- **Our earlier claim that "`pack:` content is language-scoped elsewhere" was wrong.** A concrete counterexample: `Chronicles Foreign Black Border Japanese Booster Pack` (BCHR) has its `pack:` content pointing at `{set: chr, code: default}` — the *English* CHR set's own booster/sheet definition — sharing the exact same card uuids as the English CHR product, for the same underlying reason `deck:` content does. The difference is structural: BCHR is a wholly separate top-level MTGJSON/Scryfall set (linked back to `chr` via `Set.parentCode`/`isForeignOnly`, confirmed against Scryfall's own `parent_set_code: chr` for BCHR), not a `language`-tagged row on the same set, so resolving it needs cross-set card matching (e.g. by collector number against the `parentCode`d set), not the same-set `cards_by_language` lookup this PR adds for `deck:`. Left as a follow-up; flagging here so it isn't lost.

## Test plan

- [x] `ruff check` / `ruff format --check` on all touched files
- [x] `pylint` 10.00/10 on all touched files
- [x] `mypy --explicit-package-bases` clean on all touched files
- [x] Full existing test suite passes unchanged (`pytest tests/ --record-mode=none`)
- [x] New tests in `tests/mtgjson5/test_derived_foreign_data_source_products.py`: foreign entries get their own `sourceProducts` distinct from the parent card's, unmatched/empty/`None`-`card_to_products_df` cases, row order/identity preserved across multiple cards, accepts both eager and lazy frames
- [x] New tests in `tests/mtgjson5/test_sealed_deck_language.py`: an untagged product sharing a deck keeps the deck's own uuids (regression guard for the Junji Ito English/Japanese case sharing one deck object), a language-tagged product substitutes the foreign uuid per card, finish is still resolved from the original uuid, missing-foreign-entry fallback warns instead of dropping/crashing, unknown language/deck name handled gracefully


kodawah:
tldr: we need to decide if we're ok having the decklist and booster content list non-english uuid (the ones present in foreignData) or if the switcharoo should happen in the consumer of this data

